PR #7, opened 2020-08-17
improve error messages and docs

### LINKED ISSUE #5: STM32F407VGTx Do not have unwind info for the given address
Do we think its a configuration error or unsupported arch somewhere?

```bash
$ cargo run --release --example cycle_count
    Finished release [optimized] target(s) in 0.01s
     Running `probe-run --chip STM32F407VGTx target/thumbv7em-none-eabi/release/examples/cycle_count`
flashing program ..
DONE
resetting device
ticks: 16800059
stack backtrace:
   0: 0x0800132e - <unknown>
Error: Do not have unwind info for the given address.
```

japaric:
closes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ manager before installing `probe-run`.
 
 ## Setup
 
+1. Set the Cargo runner
+
 The recommend way to use `probe-run` is to set as the Cargo runner of your application.
 Add this line to your Cargo configuration (`.cargo/config`) file:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,6 +44,30 @@ Instead of `$CHIP` you'll need to write the name of your microcontroller.
...
+If you are using the `cortex-m-quickstart` template then this is already the case.
+If not check or add these lines to `Cargo.toml`.
+
+``` toml
+# Cargo.toml
+[profile.dev]
+debug = 1 # default is `true`; not needed if not already overridden
+
+[profile.release]
+debug = 1 # default is `false`; using `true` is also OK
```
jonas-schievink, line 59: Why do we need to enable this? Backtraces should work even without line info

japaric: the virtual unwinder (gimli) requires `.debug_frame` info to compute the stack pointer of the previous frame while walking up the stack. Last time I checked `.debug_frame` is not produced when `debug = false`.

jonas-schievink: Hmm, then how does it work as the backtrace implementation of libstd? That one works even without any debuginfo in the program.

japaric: Hmm, it seems `.debug_frame` is produced even with `debug = false`. Setting `debug = false` for `profile.dev` makes `probe-run` unwinder fail though; may be due to some other reason.

jonas-schievink: Okay, seems fine to merge this then. Maybe file an issue about the unwinder bug?

japaric: #9

PR #11, opened 2020-08-18
FPU register unstacking support

### LINKED ISSUE #1: virtual unwind support for hard-float targets
general comment: the Thumb targets are backwards compatible. Both `thumbv7em-none-eabi` and `thumbv7em-none-eabihf` work fine on Cortex-M4F devices. If you are not doing floating point operations (`f32` math) then you can use the `thumbv7em-none-eabi` target, which is supported by `probe-run`, without any performance degradation.

---

The virtual unwinder doesn't expect floating pointer (FP) registers to be pushed on exception entry.

The Cortex-M is smart about this: it only pushes FP registers if the preempted context was using them.
The unwinder will need to compute at runtime whether the FP registers were pushed or not (there should be some flag/register pushed onto the stack that contains this info) and update the per-frame SP value accordingly while walking up the stack.

jonas-schievink:
Fixes #1 

cc @mattico, would be great if you could test this out

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,11 +29,7 @@ use probe_rs::{
 };
 use probe_rs_rtt::{Rtt, ScanRegion, UpChannel};
 use structopt::StructOpt;
-use xmas_elf::{
-    header::HeaderPt2, program::Type, sections::SectionData, symbol_table::Entry, ElfFile,
-};
-
-const EF_ARM_ABI_FLOAT_HARD: u32 = 0x00000400;
+use xmas_elf::{program::Type, sections::SectionData, symbol_table::Entry, ElfFile};
 
 const TIMEOUT: Duration = Duration::from_secs(1);
 
@@ -98,18 +94,6 @@ fn notmain() -> Result<i32, anyhow::Error> {
             }
         })
         .next();
-
-    let flags = match elf.header.pt2 {
-        HeaderPt2::Header32(p32) => p32.flags,
-        HeaderPt2::Header64(p64) => p64.flags,
-    };
-
-    if flags & EF_ARM_ABI_FLOAT_HARD != 0 {
-        bail!(
-            "the hard-float ABI is not supported.\n\
-workaround: use the `thumbv7em-none-eabi` compilation target (note: no `hf`)"
-        );
-    }
 
     let mut debug_frame = None;
     let mut range_names = None;
@@ -512,7 +496,13 @@ fn backtrace(
             return Ok(top_exception);
         }
 
-        if lr > 0xffff_fff0 {
+        if lr > 0xffff_ffe0 {
+            let fpu = match lr {
+                0xFFFFFFF1 | 0xFFFFFFF9 | 0xFFFFFFFD => false,
+                0xFFFFFFE1 | 0xFFFFFFE9 | 0xFFFFFFED => true,
+                _ => bail!("LR contains invalid EXC_RETURN value 0x{:08X}", lr),
+            };
+
             // we walk the stack from top (most recent frame) to bottom (oldest frame) so the first
             // exception we see is the top one
             if top_exception.is_none() {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -512,7 +496,13 @@ fn backtrace(
             return Ok(top_exception);
         }
 
-        if lr > 0xffff_fff0 {
+        if lr > 0xffff_ffe0 {
+            let fpu = match lr {
+                0xFFFFFFF1 | 0xFFFFFFF9 | 0xFFFFFFFD => false,
+                0xFFFFFFE1 | 0xFFFFFFE9 | 0xFFFFFFED => true,
```
japaric, line 502: ahh, so that's where the FPU flag is stored. Clever.

src/main.rs
```diff
@@ -576,12 +587,14 @@ struct Stacked {
...
     xpsr: u32,
+    fpu_regs: Option<StackedFpuRegs>,
 }
 
 impl Stacked {
-    fn read(core: &mut Core<'_>, sp: u32) -> Result<Self, anyhow::Error> {
-        let mut registers = [0; 8];
-        core.read_32(sp, &mut registers)?;
+    fn read(core: &mut Core<'_>, sp: u32, fpu: bool) -> Result<Self, anyhow::Error> {
+        let mut storage = [0; 8 + 17];
```
japaric, line 595: could we turn these "8" and "17" into constants? Perhaps as associated constants of `StackedFpuRegs`? Maybe even defined as `mem::size_of::<Self>()`?

japaric:
I would suggest using this program as a test:
``` rust
static X: AtomicU32 = AtomicU32::new(0);
static Y: AtomicU32 = AtomicU32::new(0);
static Z: AtomicU32 = AtomicU32::new(0);

#[entry]
fn main() -> ! {
    let x = f32::from_bits(X.load(Ordering::Relaxed));
    let y = f32::from_bits(Y.load(Ordering::Relaxed));
    SCB::set_pendsv();
    let z = x * y;
    Z.store(z.to_bits(), Ordering::Relaxed);

    loop {
        asm::bkpt()
    }
}

#[exception]
fn PendSV() {
    // prevent const propagation optimization
    X.store(1, Ordering::Relaxed);
    Y.store(1, Ordering::Relaxed);
    Z.load(Ordering::Relaxed);

    loop {
        asm::bkpt()
    }
}
```

It produces this machine code:
``` armasm
0000040a <nrf52::__cortex_m_rt_main>:
 40a:   b580            push    {r7, lr}
 40c:   466f            mov     r7, sp
 40e:   f240 0000       movw    r0, #0
 412:   f04f 5380       mov.w   r3, #268435456  ; 0x10000000
 416:   f2c2 0000       movt    r0, #8192       ; 0x2000
 41a:   6801            ldr     r1, [r0, #0]
 41c:   6842            ldr     r2, [r0, #4]
 41e:   ee00 1a10       vmov    s0, r1
 422:   ee01 2a10       vmov    s2, r2
 426:   f64e 5204       movw    r2, #60676      ; 0xed04
 42a:   f2ce 0200       movt    r2, #57344      ; 0xe000
 42e:   ee20 0a01       vmul.f32        s0, s0, s2
 432:   6013            str     r3, [r2, #0]
 434:   ee10 1a10       vmov    r1, s0
 438:   6081            str     r1, [r0, #8]
 43a:   f000 f86f       bl      51c <__bkpt>
 43e:   e7fc            b.n     43a <nrf52::__cortex_m_rt_main+0x30>
```

which to me looks like it keeps `s0` (and `s2`?) in use when PendSV is called. I think that's the required condition to trigger FPU register stacking.

@jonas-schievink it may be worth to temporarily add some log statements that report where the FPU registers were stacked or not. That way people testing this can get feedback on whether the hit the right path of the unwinder.

jonas-schievink:
I've partially tested this, but I haven't managed to create a binary that actually fails to unwind without the stacking changes, so I've only really tested the LR range check.

jonas-schievink:
> @jonas-schievink it may be worth to temporarily add some log statements that report where the FPU registers were stacked or not. That way people testing this can get feedback on whether the hit the right path of the unwinder.

Yup, this seems to work fine.

jonas-schievink:
Merging for now, if we find bugs in the logic here we can fix them later

mattico:
I can confirm that this works, Thanks!

jacobrosenthal:
Ditto! STM32F407VGTx

PR #12, opened 2020-08-18
Clean up CLI argument handling

jonas-schievink:


src/main.rs
```diff
@@ -49,24 +49,24 @@ struct Opts {
...
 
-    let opts = Opts::from_args();
+    let opts: Opts = Opts::from_args();
 
     if opts.list_chips {
         return print_chips();
     }
 
-    let bytes = fs::read(&opts.elf)?;
+    let elf_path = opts.elf.as_deref().unwrap();
```
Lotterleben, line 67: ha, nice

Lotterleben (approved):
can't test locally right now due to time constraints & a credential situation so I'll trust you that this works

PR #18, opened 2020-08-19
Add `panic-probe`

jonas-schievink:


jacobrosenthal:
Im sure you all are going to do a defmt based one eventually, but until then. I was thinking something like panic-rtt-break would be a good solution to do the print using rprintln so it would fit in over there
https://github.com/mvirkkunen/rtt-target/issues/8

panic-probe/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+mod imp {
+    use core::panic::PanicInfo;
+
+    use cortex_m::asm;
+    use cortex_m_rt::{exception, ExceptionFrame};
+
+    #[panic_handler]
+    fn panic(_: &PanicInfo) -> ! {
+        // Trigger a `HardFault`.
+        // FIXME: This will actually cause a `UsageFault` if that's enabled!
```
japaric, line 25: can we disabled `UsageFault` from here? It's OK to disable interrupts if that's a potential failure point. This is the unrecoverable error path anyways.

japaric:
> to do the print using rprintln so it would fit in over there

discussed this a bit with the PR author yesterday and I think we may want to add Cargo features to pick between printing the message using `rtt-target` and printing the message (or just the panic location) using `defmt`.

panic-probe/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+        static PANICKED: AtomicBool = AtomicBool::new(false);
+
+        // Guard against infinite recursion, just in case.
+        if PANICKED.load(Ordering::Relaxed) {
+            loop {
+                asm::bkpt();
+            }
+        }
+
+        PANICKED.store(true, Ordering::Relaxed);
```
japaric, line 56: I think you want a compile fence here otherwise the compiler is free to move this `store` to after `print` (which never refers to `PANICKED`) and that could result to `panic` being re-entered due to interrupts or panics from `print`.

japaric: Also I think we may want to disable interrupts at the start of `panic` otherwise this scenario becomes possible:
- `main` panics; `PANICKED <- true`; we enter `print`
- interrupt (e.g. timer) fires and panics; that follows the `PANICKED = true` branch and exits `probe-run`

then `probe-run` will see an incomplete log frame / panic message

panic-probe/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+        let mut sink = Sink {
+            buf: &mut buf,
+            pos: 0,
+            overflowed: false,
+        };
+        write!(sink, "{}", info).ok();
+
+        let msg = str::from_utf8(&sink.buf[..sink.pos]).unwrap_or("<utf-8 error>");
+        let overflow = if sink.overflowed { OVERFLOW_MARK } else { "" };
+        defmt::error!("panic-probe: {:str}{:str}", msg, overflow);
```
japaric, line 149: in this case it doesn't hurt to include "panic-probe" (it doesn't cost anything) but it will become a bit redundant once we add file locations to log frames

panic-probe/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+branch = "main"
+optional = true
+
+[features]
+# Print the panic message using `rtt-target`.
+print-rtt = ["rtt-target"]
+# Print the panic message using `defmt`.
+print-defmt = ["defmt", "defmt-error"]
+
+defmt-error = [] # internal feature, do not use
```
japaric, line 27: where is this Cargo feature used from?

japaric (commented):
The formatting looks good to me; I left some comment about handling re-entrancy.

panic-probe/Cargo.toml
jonas-schievink: The `print-defmt` feature enables it

japaric: ah, right this is from the `defmt` log macro; didn't explicitly see it as `cfg(feature = "defmt-error")` when I searched through diff. But that makes sense.

panic-probe/src/lib.rs
jonas-schievink: If the panic handler reenters itself, the `load` would synchronize with the `store`, so this should only be a problem when an interrupt fires and also panics.

panic-probe/src/lib.rs
jonas-schievink: (removed the prefix, the text already includes "panicked at", so it's clear that a panic has occurred at that point)

panic-probe/src/lib.rs
jonas-schievink: (I've added a call to `interrupt::disable()` though)

japaric (approved):

Issue #9, opened 2020-08-17
`release/debug=false` works but `dev/debug=false` doesn't


japaric:
If release works then dev should also work.

STR: with `defmt`'s log example (all the other examples produce similar errors)

``` console
$ # dev/debug=false
$ cargo run --bin log
(..)
stack backtrace:
   0: 0x00000bec - __bkpt
   1: 0x00000632 - log::__cortex_m_rt_main
   2: 0x000002d6 - main
   3: 0x0000261a - Reset
Error: debug information is missing. Likely fixes:
1. compile the Rust code with `debug = 1` or higher. This is configured in the `profile.*` section of Cargo.toml
2. use a recent version of the `cortex-m` crates (e.g. cortex-m 0.6.3 or newer). Check versions in Cargo.lock
3. if linking to C code, compile the C code with the `-g` flag

Caused by:
    Do not have unwind info for the given address.

$ # release/debug=false
$ cargo run --release --bin log
stack backtrace:
   0: 0x00000974 - __bkpt
   1: 0x0000080e - log::__cortex_m_rt_main
   2: 0x00000108 - main
   3: 0x000008c2 - Reset

$ # dev/debug=true
$ cargo run --bin log
stack backtrace:
   0: 0x00000ede - __bkpt
   1: 0x000009ea - log::__cortex_m_rt_main
   2: 0x0000065c - main
   3: 0x00002b2a - Reset
```

jonas-schievink:
It looks like the backtrace crate (and thus libstd) do still use libunwind to unwind the stack, just symbolication is done with gimli. So that explains why it works there even without debuginfo.

jonas-schievink:
It actually looks like there simply isn't *any* DWARF data for certain functions when setting `debug = false`. My current assumption is that this means that unwinding is trivial (ie. just follow LR), and that seems to hold up in practice so far, except that we cannot detect the end of the stack properly since 0xFFFFFFFF never ends up in LR this way.

japaric:
Hitting this with the LPC845. I get roughly the same output using both the `dev` and `release` profile:

``` text
stack backtrace:
   0: 0x000002f0 - __bkpt
   1: 0x000002e4 - my_app::exit
   2: 0x00000160 - hello::__cortex_m_rt_main
   3: 0x000000c8 - main
   4: 0x000001a2 - Reset
   5: 0x0f0002e4 - <unknown>
```

[...]

PR #23, opened 2020-08-19
Simplify obtaining unwind info

jonas-schievink:
Found while investigating #9, but this doesn't seem to change behavior at all

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -471,13 +471,12 @@ fn backtrace(
             .unwrap_or("<unknown>");
         println!("{:>4}: {:#010x} - {}", frame, pc, name);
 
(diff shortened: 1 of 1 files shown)


japaric (approved):
I felt like there was some slight difference between FDE and unwind info (`.debug_frame` vs `.debug_unwind`?) but if backtraces across external assembly like `__bkpt` still work then ship it.

jonas-schievink:
Looks like the implementation just does what we did before:

https://github.com/gimli-rs/gimli/blob/9d2c8f30760dac07461b8cfc0cc7e614cc9fa5dd/src/read/cfi.rs#L387-L388

PR #24, opened 2020-08-19
Fix unwinding trivial frames without debug info

### LINKED ISSUE #9: `release/debug=false` works but `dev/debug=false` doesn't

jonas-schievink:
Fixes #9 (hopefully)

This needs more testing, and the FIXME should probably be addressed as well.

Previously we couldn't produce this backtrace for the `exception` defmt example when building with `debug = 0`:

```
stack backtrace:
   0: 0x000016fc - __bkpt
   1: 0x00000301 - exception::__cortex_m_rt_PendSV
   2: 0x00000265 - PendSV
      <exception entry>
   3: 0x00001688 - core::ptr::write_volatile
   4: 0x00000439 - cortex_m::peripheral::scb::<impl cortex_m::peripheral::SCB>::set_pendsv
   5: 0x000001c5 - exception::__cortex_m_rt_main
   6: 0x0000012b - main
   7: 0x00001e37 - Reset
```

We'd get stuck at `core::ptr::write_volatile`, since that does not have a Frame Description Entry (FDE). However, that function is trivial, preserves LR, and uses no stack, so we don't need any further info to unwind past it.

This did not occur with `debug = 1` (or higher), since that apparently causes additional instructions to be emitted for `write_volatile`, which in turn makes it need an FDE to properly unwind. It also tends to go away when optimizations are enabled, presumably because (often) no trivial functions are left over after inlining.

(this is my current theory of what goes wrong and could use some more confirmation)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -471,13 +471,12 @@ fn backtrace(
             .unwrap_or("<unknown>");
         println!("{:>4}: {:#010x} - {}", frame, pc, name);
 
-        let fde = debug_frame.fde_for_address(bases, pc.into(), DebugFrame::cie_from_offset).with_context(|| {
+        let uwt_row = debug_frame.unwind_info_for_address(bases, ctx, pc.into(), DebugFrame::cie_from_offset).with_context(|| {
             "debug information is missing. Likely fixes:
 1. compile the Rust code with `debug = 1` or higher. This is configured in the `profile.*` section of Cargo.toml
 2. use a recent version of the `cortex-m` crates (e.g. cortex-m 0.6.3 or newer). Check versions in Cargo.lock
 3. if linking to C code, compile the C code with the `-g` flag"
         })?;
-        let uwt_row = fde.unwind_info_for_address(&debug_frame, bases, ctx, pc.into())?;
 
         let cfa_changed = registers.update_cfa(uwt_row.cfa())?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,7 +15,7 @@ use std::{
     time::Duration,
 };
 
-use anyhow::{anyhow, bail, Context};
+use anyhow::{anyhow, bail};
 use arrayref::array_ref;
 use gimli::{
     read::{CfaRule, DebugFrame, UnwindSection},
@@ -457,6 +457,10 @@ fn backtrace(
     let mut registers = Registers::new(lr, sp, core);
     println!("stack backtrace:");
     loop {
+        // FIXME: Add some sanity check for the PC value to avoid unwinding a corrupted stack.
+        // For example, PC should always have the LSb (thumb bit) set, and should be located within
+        // `.text` (it is possible to execute code from RAM dynamically, but we cannot unwind that).
+
         let name = range_names
             .binary_search_by(|rn| {
                 if rn.0.contains(&pc) {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -457,6 +457,10 @@ fn backtrace(
     let mut registers = Registers::new(lr, sp, core);
     println!("stack backtrace:");
     loop {
+        // FIXME: Add some sanity check for the PC value to avoid unwinding a corrupted stack.
+        // For example, PC should always have the LSb (thumb bit) set, and should be located within
+        // `.text` (it is possible to execute code from RAM dynamically, but we cannot unwind that).
```
japaric, line 462: (it theory `.text` can be placed in RAM and it may or may not be modified on the fly.)

src/main.rs
```diff
@@ -471,27 +475,28 @@ fn backtrace(
...
+            cfa_changed = registers.update_cfa(uwt_row.cfa())?;
 
-        for (reg, rule) in uwt_row.registers() {
-            registers.update(reg, rule)?;
+            for (reg, rule) in uwt_row.registers() {
+                registers.update(reg, rule)?;
+            }
         }
 
+        // We might not find unwind info for some PC values. In that case, we assume that unwinding
```
japaric, line 489: "PC values with no unwind info" could be external assembly with no CFI instructions or external C not compiled with the `-g` flag (I think). Those may not have trivial unwinding.

japaric (commented):
shared some thoughts haven't look at the code in detail

jonas-schievink:
Hmm, it seems that this doesn't fix the `panic` example, so now that gets a truncated backtrace without probe-run exiting with an error code or printing any error:

```
stack backtrace:
   0: 0x00001478 - __bkpt
   1: 0x0000195b - panic_probe::__cortex_m_rt_HardFault
```

expected output:

```
stack backtrace:
   0: 0x00001694 - __bkpt
   1: 0x00001b7e - panic_probe::__cortex_m_rt_HardFault
   2: 0x00001ffc - HardFault
      <exception entry>
   3: 0x000016a6 - __udf
   4: 0x00001b88 - cortex_m::asm::udf
   5: 0x00001b6e - rust_begin_unwind
   6: 0x00001710 - core::panicking::panic_fmt
   7: 0x000016ea - core::panicking::panic
   8: 0x000001c4 - panic::__cortex_m_rt_main
   9: 0x00000114 - main
  10: 0x00001dd6 - Reset
```

jonas-schievink:
Looks like this needs a different approach

PR #25, opened 2020-08-19
Increase RTT attach retries

Bobo1239:
With a blue pill (STM32F103C8) the number of retries wasn't sufficient to successfully attach when using a debug build. The new number works now at least for my hardware.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -331,7 +331,7 @@ fn setup_logging_channel(
     sess: Arc<Mutex<Session>>,
 ) -> Result<UpChannel, anyhow::Error> {
     if let Some(rtt_addr_res) = rtt_addr {
-        const NUM_RETRIES: usize = 5; // picked at random, increase if necessary
+        const NUM_RETRIES: usize = 10; // picked at random, increase if necessary
         let mut rtt_res: Result<Rtt, probe_rs_rtt::Error> =
             Err(probe_rs_rtt::Error::ControlBlockNotFound);
 


japaric:
The `rtt-target` API is not great here: it initializes the RTT block at runtime meaning that it can take an arbitrary amount of time between the start of the program and the RTT initialization; or worse, the RTT block may never be initialized but the static variable (RTT_BLOCK) may still be present (thus `probe-run` will hit this retry-wait loop path).
Issues depicted below.

``` rust
// issue-1.rs
delay(arbitrary_amount_of_time);
rtt_init_print!(NoBlockSkip, SIZE);
```

``` rust
// issue-2.rs
if condition_that_evaluates_to_false() {
    rtt_init_print!();
}
```

Also it seems that `rtt_init_print!` takes `O(SIZE)` time to initialize the RTT block in unoptimized builds.
We have had reports from people using unoptimized builds saying that they have to wait a long time before logs are displayed.

I don't think these conditions can be perfectly detected and avoided.
In particular, every time we make `RETRIES` larger the user-experience for `issue-2` gets worse: the user has to wait long time before they get the error message about the RTT block not being runtime initialized.

I'm going to take this PR but I consider the root issue unsolvable from the `probe-run` side.

P.S. None of these are issues with the upcoming `defmt-rtt` transport (because the RTT block is const-time initialized) so I think we should recommend that over `rtt-target` when it's publicly released.

japaric (approved):

mvirkkunen:
> The `rtt-target` API is not great here: it initializes the RTT block at runtime meaning that it can take an arbitrary amount of time between the start of the program and the RTT initialization;

It's intended for the user to call the `rtt_init_*` macros as the first thing in their program. If they purposefully decide to insert a delay before that, that's user error.

> or worse, the RTT block may never be initialized but the static variable (RTT_BLOCK) may still be present (thus `probe-run` will hit this retry-wait loop path).

That's why there is a timeout.

> Also it seems that `rtt_init_print!` takes `O(SIZE)` time to initialize the RTT block in unoptimized builds.

Yes, currently the buffer ends up in `.bss` which causes it to be zeroed at startup, but we're talking maybe milliseconds of time here in practice. `defmt` seems to be using a special section (`.uninit`) for it so the zeroing is avoided. I didn't even know a section like that existed, but I don't think it's standard. It's not in other `-rt` linked scripts I checked, and `rtt-target ` intends to support multiple platforms beyond ARM.

> We have had reports from people using unoptimized builds saying that they have to wait a long time before logs are displayed.

Long time being tens of milliseconds or tens of seconds? If latter, the reports should probably be investigated further.

> I don't think these conditions can be perfectly detected and avoided.
> In particular, every time we make `RETRIES` larger the user-experience for `issue-2` gets worse: the user has to wait long time before they get the error message about the RTT block not being runtime initialized.

There's also the bootloader case (#35) where there can be an arbitrary delay even before `defmt-rtt`'s initialization code on the application side (which piggybacks on the code that initializes the `.data` section) gets to run. That issue was reported after your comment though, but it's good to keep in mind.

> I'm going to take this PR but I consider the root issue unsolvable from the `probe-run` side.
> 
> P.S. None of these are issues with the upcoming `defmt-rtt` transport (because the RTT block is const-time initialized) so I think we should recommend that over `rtt-target` when it's publicly released.

Or we could just all try to get along and I can borrow your idea about initializing the control block via whatever initializes `.data` too. That avoids the user error.

PR #30, opened 2020-08-22
Add --no-flash option to probe-run

therealprof:
Skipping the actual flashing process is useful if the binary to be run
is already in flash, either due being a re-run or use of a different
approach to do the flashing.

Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,13 +42,13 @@ struct Opts {
     #[structopt(long)]
     list_chips: bool,
     #[cfg(feature = "defmt")]
-    #[structopt(long)]
+    #[structopt(long, conflicts_with = "no_flash")]
     defmt: bool,
     #[structopt(long, required_unless("list-chips"))]
     chip: Option<String>,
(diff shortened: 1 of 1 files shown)


japaric:
I think that using `--no-flash` with `--defmt` would be error prone. There's currently no way to check whether the passed ELF and what was last flashed on the device match and if they don't match then all `defmt` messages would be detected as corrupted. So for now (*) I would like a CLI error to be thrown if one tries to use both flags.

(*) at least until we implement a mechanism to detect that what's flashed on the device matches the ELF we are going to use to decode the `defmt` data

therealprof:
I don't think it would be a lot more error prone than using without `--defmt` but I agree that it would be a lot better if we could get some feedback from cargo passed to the runner whether the implict `cargo build` modified the target or not. I looked around but couldn't find anything in the passed on environment about that.

I can make the two options mutual exclusive for now if you still prefer that.

japaric (approved):
Thanks!

PR #33, opened 2020-08-25
Automatic canary-based stack overflow detection

### LINKED ISSUE #31: Approximated stack overflow detection via stack canaries
While it is possible to link an embedded app in a way that will detect stack overflows immediately and (mostly) reliably, few people actually do this as it requires using a linker wrapper to link the app twice. cortex-m-rt also doesn't do it by default at the moment.

However, we can still do our best to help out when no built-in stack overflow protection is used: After uploading the program, but before starting it, we can fill the space in RAM right after the data used by ELF sections with a unique pattern. Then, after the program exits (or even while it still runs), we scan the filled area and look for bytes that were changed from our pattern. If we find any changed bytes, there is a high probability that the program has used too much stack and overwrote its own data sections.

There are a few things to keep in mind here:
* When "real" stack overflow protection is used, the ELF sections will be at the *end* of RAM, and our canary won't fit anywhere (but is also unnecessary).
* This is all relying on luck and is a post-mortem technique: The program state is already undefined when we detect the stack overflow.
* We need to fill "enough" RAM with the canary to get a high chance of detecting the overflow, but only as much RAM as is needed (otherwise, even regular stack usage will trip the canary).
* It is currently unclear how this interacts with program that use the heap. It probably has to be disabled then, as the heap usually starts right after the fixed ELF sections.
* While this feature has both false positives and negatives, I believe it can be enabled by default if the message is worded right. Something like "program has used at least N bytes of stack space, collision with data segments is likely". This can be worded differently based on where in the canary the modification was detected.

jonas-schievink:
Fixes #31

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,7 @@ use gimli::{
     read::{CfaRule, DebugFrame, UnwindSection},
     BaseAddresses, EndianSlice, LittleEndian, RegisterRule, UninitializedUnwindContext,
 };
-use probe_rs::config::registry;
+use probe_rs::config::{registry, MemoryRegion};
 use probe_rs::{
     flashing::{self, Format},
     Core, CoreRegisterAddress, MemoryInterface, Probe, Session,
@@ -31,6 +31,7 @@ use structopt::StructOpt;
 use xmas_elf::{program::Type, sections::SectionData, symbol_table::Entry, ElfFile};
 
 const TIMEOUT: Duration = Duration::from_secs(1);
+const STACK_CANARY: u8 = 0xAA;
 
 fn main() -> Result<(), anyhow::Error> {
     notmain().map(|code| process::exit(code))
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
I have a repo that I think I overflow pretty easily by changing window size from 16 to 64 .
https://github.com/jacobrosenthal/dsp-discoveryf4-rust/blob/master/lab4/examples/4_11_stft_accelerometer.rs

So wanted to try this. Oddly enough I get
```bash
$ cargo run --release --example 4_11_stft_accelerometer_microfft
    Finished release [optimized + debuginfo] target(s) in 0.06s
     Running `probe-run --chip STM32F407VGTx target/thumbv7em-none-eabihf/release/examples/4_11_stft_accelerometer_microfft`
flashing program ..
DONE
resetting device
Error: RTT control block not found in target memory. Make sure RTT is initialized on the target.
```

Note I actually get the same message in `main` now so its not this PR.

jonas-schievink:
@jacobrosenthal Thanks for the report, we might need to put a delay in the RTT connection loop.

src/main.rs
```diff
@@ -652,14 +753,20 @@ fn range_names_from(
...
     let mut rtt = None;
+    let mut uses_heap = false;
     if let SectionData::SymbolTable32(entries) = sd {
         for entry in entries {
             if let Ok(name) = entry.get_name(elf) {
-                if name == "_SEGGER_RTT" {
-                    rtt = Some(entry.value() as u32);
+                match name {
+                    "_SEGGER_RTT" => rtt = Some(entry.value() as u32),
+                    "__rust_alloc" | "__rg_alloc" | "__rdl_alloc" | "malloc" if !uses_heap => {
```
japaric, line 765: isn't `__rust_alloc` an implementation detail? for now it will make do though

src/main.rs
```diff
@@ -67,6 +68,30 @@ fn notmain() -> Result<i32, anyhow::Error> {
...
     let elf = ElfFile::new(&bytes).map_err(|s| anyhow!("{}", s))?;
 
+    let target = probe_rs::config::registry::get_target_by_name(chip)?;
+
+    let mut ram_region = None;
+    for region in &target.memory_map {
+        match region {
+            MemoryRegion::Ram(ram) => {
+                if let Some(old) = &ram_region {
+                    log::debug!("multiple RAM regions found ({:?} and {:?}), stack canary will not be available", old, ram);
```
japaric, line 78: the device may have two regions but the application may only use one; this logic will exclude that kind of applications
this is ok for now but should we file a ticket to tweak the logic further to allow those apps? (not sure what that logic would look like though)

japaric (approved):
looks great! left some comments about the logic but nothing of concern

src/main.rs
jonas-schievink: Yeah it is

src/main.rs
jonas-schievink: Agreed. I've filed https://github.com/knurling-rs/probe-run/issues/39.

PR #37, opened 2020-08-27
[WIP] Add crate `probe-run-ctrl`

jonas-schievink:
(this still needs to be implemented in probe-run, but I wanted to gather feedback first)

This provides a canonical way to terminate the host's `probe-run` process with either a success and failure status. While `panic-probe` does this automatically in the case of panics and HardFaults, the `probe-run-ctrl` crate can also be used directly to abort probe-run. This makes it more convenient to write custom panic handlers etc. that want to terminate `probe-run`.

Based on #18, only the last commit is new.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
         profile: minimal
         toolchain: ${{ matrix.rust }}
         override: true
+        target: thumbv6m-none-eabi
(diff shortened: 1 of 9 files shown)


jonas-schievink:
Using semihosting for exiting probe-run seems like a better idea https://github.com/knurling-rs/probe-run/issues/120

PR #38, opened 2020-08-28
display location info

japaric:
depends on knurling-rs/defmt#148
Screenshot from 2020-08-28 15-04-48


src/main.rs
```diff
@@ -77,7 +79,18 @@ fn notmain() -> Result<i32, anyhow::Error> {
             eprintln!("warning: application may be using `defmt` but `--defmt` flag was not used");
         }
 
-        table
+        let locs = if opts.defmt {
+            let locs = elf2table::get_locations(&bytes)?;
+
+            if !table.as_ref().unwrap().is_empty() && locs.is_empty() {
+                bail!("DWARF file location info not found; compile your program with `debug = 2`")
```
jonas-schievink, line 86: I think it would be good to fall back to not printing source locations instead of exiting with an error here (and above). There might be bugs in the extraction code, or the user doesn't need/want them and doesn't want to build with `debug = 2`.

Is this maybe even a sufficient way to opt out of having the locations printed? I'm sure some users will mind that the output now uses 2 lines per log frame instead of 1, so having that seems useful.

japaric: changed to emit a warning instead of an error on (a) debug = 1 and (b) incomplete location info (bug in elf2table)

japaric:
now this depends on knurling-rs/defmt#150

jonas-schievink (approved):

PR #41, opened 2020-08-29
Allow environment variables to set chip arguments.

Hoverbear:
This small change makes allowing build tooling to configure the passed chip a bit easier. I also did `elf` because it was... well, it was there. I'd be very happy to remove that change if desired.

It also educates the user on how to use the new feature, and ways they might leverage it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -55,7 +55,7 @@ struct Opts {
     chip: Option<String>,
 
     /// Path to an ELF firmware file.
-    #[structopt(name = "ELF", parse(from_os_str), required_unless("list-chips"), env = "PROBE_RUN_ELF")]
+    #[structopt(name = "ELF", parse(from_os_str), required_unless("list-chips"))]
     elf: Option<PathBuf>,
 
     /// Skip writing the application binary to flash.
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -51,11 +51,11 @@ struct Opts {
     defmt: bool,
 
     /// The chip to program.
-    #[structopt(long, required_unless("list-chips"))]
+    #[structopt(long, required_unless("list-chips"), env = "PROBE_RUN_CHIP")]
     chip: Option<String>,
 
     /// Path to an ELF firmware file.
-    #[structopt(name = "ELF", parse(from_os_str), required_unless("list-chips"))]
+    #[structopt(name = "ELF", parse(from_os_str), required_unless("list-chips"), env = "PROBE_RUN_ELF")]
```
jonas-schievink, line 58: Is there a use case for controlling this through environment variables?

jonas-schievink (commented):
Thanks, looks like a fine addition!

src/main.rs
Hoverbear: I'm not sure! I feel like no, so let me remove it.

PR #44, opened 2020-08-31
Better diagnostic when linker script was missing

### LINKED ISSUE #42: Diagnose probable lack of linker script
It's a common problem that users forget to pass `-Tlink.x` to the linker, since this currently needs to be done in every embedded application. The result is an ELF that is missing most sections. I think we do currently catch that, but we should try to include the likely cause and a suggestion in the error message.

jonas-schievink:
Before:

```
Error: .`.defmt` section not found
```

After:

```
Error: `.text` section is missing, please make sure that the linker script was passed to the linker
```

Closes #42

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -104,6 +104,13 @@ fn notmain() -> Result<i32, anyhow::Error> {
         );
     }
 
+    elf.find_section_by_name(".text").ok_or_else(|| {
+        anyhow!(
+            "`.text` section is missing, please make sure that the linker script \
+            was passed to the linker"
+        )
+    })?;
+
     #[cfg(feature = "defmt")]
     let (table, locs) = {
         let table = elf2table::parse(&bytes)?;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -104,6 +104,13 @@ fn notmain() -> Result<i32, anyhow::Error> {
         );
     }
 
+    elf.find_section_by_name(".text").ok_or_else(|| {
+        anyhow!(
+            "`.text` section is missing, please make sure that the linker script \
+            was passed to the linker"
```
japaric, line 110: maybe also suggest looking into `.cargo/config.toml`? setting `RUSTFLAGS` also breaks linking like this

japaric (approved):
looks good to me; maybe add where to look?

eHammarstrom:
This will save people getting started some time, I can vouch for that! :laughing:

Thanks @jonas-schievink

PR #45, opened 2020-08-31
Use `object` instead of `xmas-elf`

jonas-schievink:
This removes the ability to upload code straight to RAM. It is possible to reimplement the functionality with `object`, but since no apps can currently make use of this (unless they use a custom runtime crate), I thought it was better to take it out since it isn't readily testable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,6 +573,10 @@ name = "object"
 version = "0.21.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37fd5004feb2ce328a52b0b3d01dbf4ffff72583493900ed15f22d4111c51693"
+dependencies = [
+ "flate2",
+ "wasmparser",
+]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


japaric (approved):
Looks good to me!
> no apps can currently make use of this (unless they use a custom runtime crate)

you can put the app (.text and .rodata) in RAM using the `cortex-m-rt` crate by doing something like:
```
/* link.x */
MEMORY
{
  FLASH : ORIGIN = 0x20000000, LENGTH = 128K /* <- a lie */
  RAM : ORIGIN = 0x20020000, LENGTH = 128K
}
```

jonas-schievink:
Oh, okay. Do you think I should restore this feature before merging then?

japaric:
> Do you think I should restore this feature before merging then?

it's fine to remove it since no one is using it (or knows how to use it) anyways
we can add it back when cortex-m-rt has better support for apps in RAM

Issue #46, opened 2020-09-01
stack trace infinite loop


kaspar030:
I'm just getting started using probe-run, testing a threading crate I'm building.

So (on Cortex-M), the thread stacks are populated with a frame like an exception would do, ```[r0-r4, r12, LR, PC, APSR]```.

LR is set to the location of a `cleanup()`, which does some house-keeping, then triggers the scheduler to switch away, then does a ```loop {}```, which is practically unreachable.

Now when I run something with probe-run, the resulting stack trace endlessly repeats in cleanup():

```
[...]
    Finished dev [optimized + debuginfo] target(s) in 0.01s              
     Running `probe-run --chip nRF52840_xxAA /home/kaspar/src/own/rust/riot_core/target/thumbv7em-none-eabi/debug/examples/bench_lock`             
flashing program ..                                                      
^[[CDONE                                                                 
resetting device                                                         
stack backtrace:                                                         
   0: 0x00001752 - <unknown>                                             
   1: 0x000016b6 - cortex_m_semihosting::hio::hstdout
   2: 0x00001650 - cortex_m_semihosting::export::hstdout_fmt::{{closure}}                                                                          
   3: 0x00001558 - cortex_m::interrupt::free                             
   4: 0x00001638 - cortex_m_semihosting::export::hstdout_fmt
   5: 0x000007b6 - user_main                                             
   6: 0x0000112a - riot_core::main_trampoline                            
   7: 0x00000d20 - riot_core::thread::cleanup                                                                                                      
   8: 0x00000d20 - riot_core::thread::cleanup                            
   9: 0x00000d20 - riot_core::thread::cleanup       
  10: 0x00000d20 - riot_core::thread::cleanup                            
  11: 0x00000d20 - riot_core::thread::cleanup            
  12: 0x00000d20 - riot_core::thread::cleanup
  13: 0x00000d20 - riot_core::thread::cleanup
[...]
```

The lines ```<N>: 0x00000d20 - riot_core::thread::cleanup``` repeat extremely fast, and the process cannot be cancelled with CTRL-C.

While I might have messed up something on the stack to make it invalid and look like this to probe-run, I'd expect it to stop after maybe the second iteration, letting the user know that this would repeat.

(I'm assuming that probe-run's rtt support uses a different BKPT scheme than traditional semihosting, and that's why this exits on the first ```hprintln!()```, but that is a unrelated to this issue.)

This is on probe-run 0.1.3 installed today via "cargo install probe-run".

[...]

PR #47, opened 2020-09-01
ctrlc -> signal-hook, remove the handler when done

jonas-schievink:
This allows canceling the backtrace printing, which is sometimes broken and loops infinitely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,6 +46,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6b602bfe940d21c130f3895acd65221e8a61270debe89d628b9cb4e3ccb8569b"
 
 [[package]]
+name = "arc-swap"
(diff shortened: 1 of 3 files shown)


jonas-schievink:
This seems to work fine, merging to help with #46

kaspar030:
> This allows canceling the backtrace printing, which is sometimes broken and loops infinitely.

This still does not work for me. When doing "cargo run" or directly "probe-run ..." for testing #46, ctrl-c doesn't have any effect.

jonas-schievink:
Ah, sorry, I should have properly read the docs of [`signal_hook::unregister`](https://docs.rs/signal-hook/0.1.16/signal_hook/fn.unregister.html)

jonas-schievink:
Opened https://github.com/knurling-rs/probe-run/pull/51 to fix this

PR #49, opened 2020-09-01
Allow listing and selecting probe

### LINKED ISSUE #14: add flag / env-var to select the probe by serial number
This is helpful when you want to `cargo-run` firmware on two or more different devices in parallel.

``` console
$ # terminal 1
$ cd path/to/firmware
$ PROBE_RUN_SN=abcdef012345 cargo run --bin radio-tx

$ # terminal 2
$ cd path/to/firmware
$ PROBE_RUN_SN=012345abcdef cargo run --bin radio-rx
```

In the above case the firmware can live in the same Cargo project.

When the firmware lives in different Cargo projects you can use the flag version.

``` console
$ # terminal 1
$ cd path/to/firmware-a
$ head .cargo/config
[target.thumbv7em-none-eabihf]
runner = "probe-run --serial-number abcdef012345 --chip nrf52"

$ cargo run --bin radio-tx

$ # terminal 2
$ cd path/to/firmware-b
$ head .cargo/config
[target.thumbv7em-none-eabihf]
runner = "probe-run --serial-number 012345abcdef --chip nrf52"

$ cargo run --bin radio-rx
```

aurelj:
This provides options similar to cargo-flash to list all connected probes
and to specify which one to use:
  probe-run --list-probes
  probe-run --probe '1366:0101' --chip nrf52 bin
  probe-run --probe '1366:0101:123456' --chip nrf52 bin

This also alows specifying the probe as en env-var:
  PROBE_RUN_PROBE='1366:0101:123456' cargo run

It fixes #14.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,17 @@ To support multiple devices, or permit overriding default behavior, you may pref
 `${PROBE_RUN_CHIP}` environment variable, and set `runner` (or
 `CARGO_TARGET_${TARGET_ARCH}_RUNNER`) to `probe-run`.
 
+If you have several probes connected, you can specify which one to use by adding
+the --probe option to the `runner` or setting the `${PROBE_RUN_PROBE}` environment
+variable with a value containing either `${VID}:${PID}` or `${VID}:${PID}:${SERIAL}`:
+
+```console
+probe-run --probe '0483:3748' --chip ${PROBE_RUN_CHIP}
+PROBE_RUN_PROBE='1366:0101:123456' cargo run
+```
+
+To list all connected probes, run `probe-run --list-probes`.
+
 2. Enable debug info
 
 Next check that debug info is enabled for all profiles.
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,7 +30,7 @@ use object::{
 use probe_rs::config::{registry, MemoryRegion};
 use probe_rs::{
     flashing::{self, Format},
-    Core, CoreRegisterAddress, MemoryInterface, Probe, Session,
+    Core, CoreRegisterAddress, DebugProbeInfo, DebugProbeSelector, MemoryInterface, Probe, Session,
 };
 use probe_rs_rtt::{Rtt, ScanRegion, UpChannel};
 use structopt::StructOpt;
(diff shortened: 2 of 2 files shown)


src/main.rs
```diff
@@ -221,7 +233,12 @@ fn notmain() -> Result<i32, anyhow::Error> {
         bail!("no probe was found")
     }
     log::debug!("found {} probes", probes.len());
-    let probe = probes[0].open()?;
+    let probe = if let Some(probe_opt) = opts.probe.as_deref() {
+        let selector: probe_rs::DebugProbeSelector = probe_opt.try_into()?;
+        probe_rs::Probe::open(selector)?
+    } else {
+        probes[0].open()?
```
japaric, line 240: could you change this to throw an error if there's more than one probe? the error should say something like "more than one probe found; use `--probe` to specify which one to use"

README.md
```diff
@@ -44,6 +44,16 @@ To support multiple devices, or permit overriding default behavior, you may pref
 `${PROBE_RUN_CHIP}` environment variable, and set `runner` (or
 `CARGO_TARGET_${TARGET_ARCH}_RUNNER`) to `probe-run`.
 
+If you have several probes connected, you can specify which one to use adding
```
japaric, line 47: could you add a note here indicating the syntax for the `--probe` flag? or say that the syntax can be found in the output of the `-h` command

japaric (commented):
Thanks for the PR! This looks great. I left some comments.

japaric:
Additional question: does the selector accept just the serial number (e.g. `--probe 123456`)? I think that the serial number is usually unique enough to select a probe.

src/main.rs
```diff
@@ -221,7 +233,12 @@ fn notmain() -> Result<i32, anyhow::Error> {
         bail!("no probe was found")
     }
     log::debug!("found {} probes", probes.len());
-    let probe = probes[0].open()?;
+    let probe = if let Some(probe_opt) = opts.probe.as_deref() {
+        let selector: probe_rs::DebugProbeSelector = probe_opt.try_into()?;
+        probe_rs::Probe::open(selector)?
```
japaric, line 238: ah, it seems that if you pass "$VID:$PID" as the selector and there are two connected probes with that VID PID pair then this will pick the "first one" of those two -- calling `probe-run` a second time results in a probe-rs error ("probe could not be created"). In that scenario we should throw an error and ask the user to narrow down the selection using the serial number.

Is it possible to use `DebugProbeSelector` to filter a list of probes?

src/main.rs
aurelj: Yes, good idea.

README.md
aurelj: OK, done.

src/main.rs
aurelj: Right.

I don't think probe-rs provides a way to use a `DebugProbeSelector` to filter a list of probes. But it is easy enough to implement.

Updated version now requires that only 1 probe matches the `DebugProbeSelector` to continue.

aurelj:
> Additional question: does the selector accept just the serial number (e.g. `--probe 123456`)? I think that the serial number is usually unique enough to select a probe.

No, in `DebugProbeSelector`, the `vendor_id` and `product_id` are not optional.

README.md
```diff
@@ -44,6 +44,17 @@ To support multiple devices, or permit overriding default behavior, you may pref
...
```
jonas-schievink, line 47:
```suggestion
If you have several probes connected, you can specify which one to use by adding
```

README.md
```diff
@@ -44,6 +44,17 @@ To support multiple devices, or permit overriding default behavior, you may pref
...
+the --probe option to the `runner` or setting the `${PROBE_RUN_PROBE}` environment
+variable with a value containing either `${VID}:${PID}` or `${VID}:${PID}:${SERIAL}`:
+
+```console
+probe-run --probe '0483:3748' --chip ${PROBE_RUN_CHIP}
+PROBE_RUN_PROBE='1366:0101:123456' cargo run
+```
+
+To list all connected probes run `probe-run --list-probes`.
```
jonas-schievink, line 56:
```suggestion
To list all connected probes, run `probe-run --list-probes`.
```

README.md
aurelj: OK, done.

README.md
aurelj: OK, done.

japaric (approved):
This is fantastic. Thank you!

PR #50, opened 2020-09-02
Ignore thumb bit when checking for repeated frames

### LINKED ISSUE #46: stack trace infinite loop

jonas-schievink:
Hopefully fixes #46

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,7 @@ use structopt::StructOpt;
 
 const TIMEOUT: Duration = Duration::from_secs(1);
 const STACK_CANARY: u8 = 0xAA;
+const THUMB_BIT: u32 = 1;
 
 fn main() -> Result<(), anyhow::Error> {
     notmain().map(|code| process::exit(code))
@@ -582,7 +583,10 @@ fn backtrace(
             break;
         }
 
-        if !cfa_changed && lr == pc {
+        // If the frame didn't move, and the program counter didn't change, bail out (otherwise we
+        // might print the same frame over and over).
+        // Since we strip the thumb bit from `pc`, ignore it in this comparison.
+        if !cfa_changed && lr & !THUMB_BIT == pc & !THUMB_BIT {
             println!("error: the stack appears to be corrupted beyond this point");
             return Ok(top_exception);
         }
@@ -616,7 +620,7 @@ fn backtrace(
             if lr & 1 == 0 {
                 bail!("bug? LR ({:#010x}) didn't have the Thumb bit set", lr)
             }
-            pc = lr & !1;
+            pc = lr & !THUMB_BIT;
         }
 
         frame += 1;


kaspar030:
This fixes it:

```
    Finished dev [optimized + debuginfo] target(s) in 2.59s
     Running `probe-run --chip nRF52840_xxAA target/thumbv7em-none-eabi/debug/app`
flashing program ..
DONE
resetting device
RTT logs not available; blocking until the device halts..
stack backtrace:
   0: 0x00000946 - <unknown>
   1: 0x000008aa - cortex_m_semihosting::hio::hstdout
   2: 0x000007ec - cortex_m_semihosting::export::hstdout_str::{{closure}}
   3: 0x0000071e - cortex_m::interrupt::free
   4: 0x000007d8 - cortex_m_semihosting::export::hstdout_str
   5: 0x0000045a - user_main
   6: 0x00001296 - riot_core::main_trampoline
   7: 0x00000fb2 - riot_core::thread::cleanup
error: the stack appears to be corrupted beyond this point
```

src/main.rs
```diff
@@ -581,7 +581,7 @@ fn backtrace(
             break;
         }
 
-        if !cfa_changed && lr == pc {
+        if !cfa_changed && lr & !1 == pc & !1 {
```
kaspar030, line 584: personally I'd expect a comment on thy the lsb is ignored. :)

japaric, line 584: can we move the `1` constant into a `const THUMB_BIT` or similar?

japaric (approved):
Looks good to me but let's add a `const`

PR #51, opened 2020-09-02
Actually deregister the Ctrl+C handler

jonas-schievink:


kaspar030:
This let's me correctly break out of the loop in #46 using ctrl-c!

japaric (approved):
(I thought we were using the `ctrlc` crate) Not familiar with this crate but I trust you.

jonas-schievink:
I switched to `signal_hook` since `ctrlc` does not support deregistering the handler at all

PR #52, opened 2020-09-03
Update probe-rs to 0.9.0

jonas-schievink:
Seems to work without code changes.


jonas-schievink:
Verdict: Having to manually install libftdi on Windows is an unacceptable usability regression IMO, so we'll wait until that works automagically.

jonas-schievink:
https://github.com/probe-rs/probe-rs/pull/378 makes libftdi an optional feature, so we can update once the next version of probe-rs is published

jonas-schievink:
probe-rs has been updated in https://github.com/knurling-rs/probe-run/pull/108

PR #57, opened 2020-09-07
print module path

japaric:
depends on knurling-rs/defmt#167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,9 +161,9 @@ dependencies = [
 
 [[package]]
 name = "cortex-m-rt"
-version = "0.6.12"
+version = "0.6.13"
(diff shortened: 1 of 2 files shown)


japaric:
I think we may want to format the module path in a separate line, right align it or something because right now it makes the start column of the log message uneven:
``` text
0.000000 INFO  lpc_app::foo::baz | hello
└─ src/foo.rs:2
0.000001 INFO  lpc_app::bar::quux | world
└─ src/bar.rs:2
0.000002 INFO  hello::__cortex_m_rt_main | Hello, world!
└─ src/bin/hello.rs:10
```

jonas-schievink (approved):

PR #58, opened 2020-09-08
stack overflow detection: adjust for full descending stacks

japaric:
in the ARM ISA the stack is of the full descending type, meaning that the
initial stack pointer will appear to be "outside" the RAM region: that is start
at `ORIGIN(RAM) + LENGTH(RAM)`
this is particularly the case when the stack overflow protection has not been
added (the stack is not flipped)
so adjust the logic accordingly

fixes: `probe-run` was printing the warning message shown below when the program panicked
``` text
  (HOST) WARN  probe_run | no RAM region appears to contain the stack; cannot determine if this was a stack overflow
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -236,8 +236,15 @@ fn notmain() -> Result<i32, anyhow::Error> {
         .memory_map
         .iter()
         .filter_map(|region| match region {
-            MemoryRegion::Ram(region) if region.range.contains(&vector_table.initial_sp) => {
-                Some(region)
+            MemoryRegion::Ram(region) => {
+                // NOTE stack is full descending; meaning the stack point can be `ORIGIN(RAM) +
+                // LENGTH(RAM)`
+                let range = region.range.start..=region.range.end;
+                if range.contains(&vector_table.initial_sp) {
+                    Some(region)
+                } else {
+                    None
+                }
             }
             _ => None,
         })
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -236,8 +236,15 @@ fn notmain() -> Result<i32, anyhow::Error> {
         .memory_map
         .iter()
         .filter_map(|region| match region {
-            MemoryRegion::Ram(region) if region.range.contains(&vector_table.initial_sp) => {
-                Some(region)
+            MemoryRegion::Ram(region) => {
+                // NOTE stack is full descending; meaning the stack point can be `ORIGIN(RAM) +
```
jonas-schievink, line 240:
```suggestion
                // NOTE stack is full descending; meaning the stack pointer can be `ORIGIN(RAM) +
```

src/main.rs
```diff
@@ -628,7 +635,10 @@ fn backtrace(
                 println!("      <exception entry>");
 
                 let stack_overflow = if let Some(sp_ram_region) = sp_ram_region {
-                    !sp_ram_region.range.contains(&sp)
+                    // NOTE stack is full descending; meaning the stack point can be `ORIGIN(RAM) +
```
jonas-schievink, line 638:
```suggestion
                    // NOTE stack is full descending; meaning the stack pointer can be `ORIGIN(RAM) +
```

jonas-schievink (approved):

PR #63, opened 2020-09-16
add location info to the backtrace

### LINKED ISSUE #17: Decorate backtrace with location information (file+line)
The libstd implementation prints file and line number for each (non-inlined) call in the backtrace. We should do the same, as it makes finding the relevant locations easier.

If the format is right, VS Code should even make the locations clickable.

japaric:
closes #17

backtrace


src/main.rs
```diff
@@ -613,24 +633,69 @@ fn backtrace(
...
+        // `find_frames` returns a wrong answer, instead of an `Err`or, when the input is the PC of
+        // a subroutine that has no debug information (e.g. external assembly). The wrong answer is
+        // one of the subroutines GC-ed by the linker so we check that the last frame
+        // (the non-inline one) is actually "live" (exists in the final binary). If it doesn't then
+        // we probably asked about something with no debug info. In that scenario we fallback to
+        // the symtab to at least provide the function name that contains the PC.
+        let subroutine = frames
+            .last()
+            .expect("BUG: `addr2line::FrameIter` was empty");
+        let has_valid_debuginfo = if let Some(function) = subroutine.function.as_ref() {
```
japaric, line 654: without this workaround `addr2line` was reporting `__udivmodsi4` as the first frame of the backtrace.

src/main.rs
```diff
@@ -228,7 +222,30 @@ fn notmain() -> Result<i32, anyhow::Error> {
...
-    let (range_names, rtt_addr, uses_heap) = range_names_from(&elf, text.index())?;
+    let text = elf
+        .section_by_name(".text")
+        .map(|section| section.index())
+        .ok_or_else(|| {
+            anyhow!(
+            "`.text` section is missing, please make sure that the linker script was passed to the \
+            linker (check `.cargo/config.toml` and the `RUSTFLAGS` variable)"
+        )
+        })?;
```
jonas-schievink, line 233: Hmm, if this is done after `elf2table::parse` we might not reach this if the linker script is missing, I think.

src/main.rs
```diff
@@ -613,24 +633,69 @@ fn backtrace(
...
+        };
+
+        if has_valid_debuginfo {
+            for frame in &frames {
+                let name = frame
+                    .function
+                    .as_ref()
+                    .map(|function| function.demangle())
+                    .transpose()?
+                    .unwrap_or(Cow::Borrowed("???"));
```
jonas-schievink, line 667: Might be better to fall back to the mangled name instead, that way the user will at least see *something*

src/main.rs
```diff
@@ -613,24 +633,69 @@ fn backtrace(
...
-            })
-            .map(|idx| &*range_names[idx].1)
-            .unwrap_or("<unknown>");
-        println!("{:>4}: {:#010x} - {}", frame, pc, name);
+            }
+        } else {
+            // .symtab fallback
+            // confusingly enough the addresses in the `.symtab` do have their thumb bit set to 1
+            // so set it back before the lookup
+            let address = (pc | THUMB_BIT) as u64;
```
jonas-schievink, line 691: This sets the thumb bit, but the comment says that it is already set here

jonas-schievink (commented):
Very nice!

src/main.rs
japaric: good point! I'll move it back to its original place.

src/main.rs
japaric: Good idea. there's also a "best effort" demangle function I think, which may help with symbols that come from C code. It could also be the case that `function` is `None` (no function name info at all); in that case we would have to print the "???"

src/main.rs
japaric: ah, the address ranges in the symtab have their thumb bits set (e.g. `0x101..0x200`) I think. `pc` has its thumb bit cleared (e.g. `0x100`) at this point; what I have seen is that the lookup can report the previously symbol if we don't set the thumb bit of `pc` before lookup. I'll reword the comment

jonas-schievink: Oh, right, that makes sense

src/main.rs
japaric: the `demangle` method [is already using the best effort `demangle_auto`](https://docs.rs/addr2line/0.13.0/src/addr2line/lib.rs.html#1308) so I think this will already fall back to the raw name if Rust-style demangling fails.
the `???` path will be followed if there's no function name info at all `frame.function` is `None`

jonas-schievink: Okay, sounds good

jonas-schievink (approved):

PR #69, opened 2020-09-21
toggle RTT mode after RAM initialization

japaric:
depends on knurling-rs/defmt#174
closes knurling-rs/defmt#133

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -333,7 +333,11 @@ fn notmain() -> Result<i32, anyhow::Error> {
 
         log::debug!("starting device");
         if core.get_available_breakpoint_units()? == 0 {
-            bail!("device without HW breakpoints are not supported");
+            if rtt_addr.is_some() {
+                bail!("RTT not supported on device without HW breakpoints");
+            } else {
+                log::warn!("device doesn't support HW breakpoints; HardFault will NOT make `probe-run` exit with an error code");
+            }
         }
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -333,10 +333,19 @@ fn notmain() -> Result<i32, anyhow::Error> {
 
         log::debug!("starting device");
         if core.get_available_breakpoint_units()? == 0 {
-            log::warn!("device doesn't support HW breakpoints; HardFault will NOT make `probe-run` exit with an error code");
-        } else {
-            core.set_hw_breakpoint(vector_table.hard_fault & !THUMB_BIT)?;
+            bail!("device without HW breakpoints are not supported");
```
japaric, line 336: FIXME: we should only bail if `rtt_addr` is `Some`

src/main.rs
```diff
@@ -933,6 +943,7 @@ fn rtt_and_heap_info_from(
         };
 
         match name {
+            "main" => main = Some(symbol.address() as u32 & !THUMB_BIT),
```
japaric, line 946: I don't think this symbol is guaranteed to exist in the final binary so we may want to do something else

jonas-schievink: At least when using cortex-m-rt it will always exist, I think

jonas-schievink (commented):
Implementation looks good, would be good to resolve the FIXME before merging though

PR #82, opened 2020-10-23
use latest git version of decoder and elf parser

japaric:


jonas-schievink:
I think we should still pin it to a git commit, because otherwise any breaking change in defmt means that probe-run can't be installed by users anymore.

japaric:
good point! pinned git deps to latest sha

jonas-schievink (approved):

PR #88, opened 2020-10-29
Added option to specify probe clock frequency

eboskma:
This adds an optional `--speed <frequency>` option to set the probe clock frequency.

I had issues flashing my Robotdyn STM32F103C8 Black Pill with my J-Link EDU Mini using any probe-rs related method (probe-run, cargo-flash, cargo-embed). When I lower the probe's clock frequency it works, but probe-run did not have the option to set the clock speed. With this PR it does!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -73,6 +73,10 @@ struct Opts {
     #[structopt(long, env = "PROBE_RUN_PROBE")]
     probe: Option<String>,
 
+    /// The probe clock frequency in kHz
+    #[structopt(long)]
+    speed: Option<u32>,
+
     /// Path to an ELF firmware file.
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Im going to need this for the samd chips as well

japaric (approved):
Thanks for the PR!

eboskma:
@japaric Thank you for your work on embedded Rust!

jacobrosenthal:
Hey @eboskma Can you try cargo flash from git? I think it has the fix for your speed issue https://github.com/probe-rs/probe-rs/pull/450
`cargo install --git https://github.com/probe-rs/cargo-flash.git` and if so would soon make its way into probe-run as well

Please let me know, were trying to track down anyone using reduced speed and get fixes in

eboskma:
@jacobrosenthal That works great, thanks for the heads up!

PR #90, opened 2020-10-30
Updated defmt to match repo

dhylands:
In particular, this addresses the issue I was seeing over here:
https://github.com/knurling-rs/app-template/issues/18


tafia:
Thanks, I just got hit by this too and I confirm installing your branch works.

Just curious, why is all of it relying on git instead of crates.io versions?

japaric:
thanks for the PR! I have updated the git deps to latest in PR #98 so I'll close this one.
`probe-run --version` now prints the supported defmt version (git hash) so that should help but we need docs on how to use that info
@tafia these dependencies are not yet on crates.io

PR #93, opened 2020-11-03
Update probe-rs to 0.9.0

kevswims:
With this update I was able to run the hello-world from the app template on an STM32G4 which was not supported in probe-rs 0.8.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ checksum = "1b6a2d3371669ab3ca9797670853d61402b03d0b4b9ebf33d677dfa720203072"
 dependencies = [
  "cpp_demangle",
  "fallible-iterator",
- "gimli 0.22.0",
+ "gimli",
  "object 0.20.0",
  "rustc-demangle",
  "smallvec",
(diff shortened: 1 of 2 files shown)


japaric:
I think we have been delaying updating this dependency because probe-rs 0.9 adds a (non-optional) C library (?) dependency that makes installation harder on Windows (?) iirc. cc @jonas-schievink

jonas-schievink:
Yes, this PR is a duplicate of https://github.com/knurling-rs/probe-run/pull/52, and CI is failing because of the missing dependency.

I'm thinking that it might be pragmatic to publish a temporary probe-rs fork so that we can update, as this seems to hit a lot of users.

japaric:
(just saw #52)

@jonas-schievink do we need a fork though? your ftdi PR has been merged so we could switch to a pinned git "version" (some recent commit) until probe-rs v0.9.1 is released

jonas-schievink:
Yeah that would work, but prevents publishing to crates.io

japaric:
> prevents publishing to crates.io

right, I was thinking of a immediate fix as we are still git only. I would be much less enthusiastic about publishing a fork of probe-rs on crates-io though (if that's what you implied)

jonas-schievink:
probe-run is already published to crates.io, but I'd be fine with using a git dependency on probe-rs temporarily.

> I would be much less enthusiastic about publishing a fork of probe-rs on crates-io though (if that's what you implied)

Yes, this is what I meant. It would of course also be a temporary solution, until probe-rs has been updated.

kevswims:
Thanks @jonas-schievink and @japaric . Should I update this to pull in a newer pinned git commit that includes the changes or close it out and wait for https://github.com/knurling-rs/probe-run/pull/52?

 As a user I would have no issue using this with a pinned git dependency since there are others.

obust:
`probe-rs` has released [0.10.0](https://github.com/probe-rs/probe-rs/blob/master/CHANGELOG.md)

jonas-schievink:
probe-rs has been updated to 0.10.0 in https://github.com/knurling-rs/probe-run/pull/108

PR #94, opened 2020-11-10
make panic-probe compile-able on hosted systems

japaric:
should unblock knurling-rs/defmt#233


jonas-schievink (commented):
This will result in an empty docs.rs page, unless we tell it to build for a different target

japaric:
set docs.rs target to thumbv7m

jonas-schievink (approved):

PR #95, opened 2020-11-10
update the CHANGELOG

japaric:
(unsure whether the next release should be 0.1.4 or v0.2.0 but in any case it will be a big release)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,74 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- [#30] added a `--no-flash` flag to run a program without re-flashing it
+[#30]: https://github.com/knurling-rs/probe-run/pull/30
+
+- [#40] added (`--help`) documentation to the many CLI flags
+[#40]: https://github.com/knurling-rs/probe-run/pull/40
+
+- [#33] added canary-based stack overflow detection
+[#33]: https://github.com/knurling-rs/probe-run/pull/33
+
+- [#38] added file location information to log messages
+[#38]: https://github.com/knurling-rs/probe-run/pull/38
+
+- [#41] the `PROBE_RUN_CHIP` env variable can be used as an alternative to the `--chip` flag
+[#41]: https://github.com/knurling-rs/probe-run/pull/41
+
+- [#49] `--list-probes` and `--probe` flags to list all probes and select a particular probe, respectively
+[#49]: https://github.com/knurling-rs/probe-run/pull/49
+
+- [#55] added more precise stack overflow detection for apps linked with `flip-link`
+[#55]: https://github.com/knurling-rs/probe-run/pull/55
+
+- [#57] added module location information to log messages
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,94 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- [#30] added a `--no-flash` flag to run a program without re-flashing it
+
+[#30]: https://github.com/knurling-rs/probe-run/pull/30
```
japaric, line 14: (I wish we had those asciidoctor macros used in the Rust-Analyzer blog post)

CHANGELOG.md
```diff
@@ -7,6 +7,94 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+
+- [#40] added (`--help`) documentation to the many CLI flags
+
+[#40]: https://github.com/knurling-rs/probe-run/pull/40
```
jonas-schievink, line 18: Interleaving them like this results in more spacing between the entries. Could you either update the new entries to not do that, or make the old entries have the same spacing?

CHANGELOG.md
```diff
@@ -7,6 +7,94 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+
+[#51]: https://github.com/knurling-rs/probe-run/pull/51
+
+- [#77] flush stdout after each write; fixes a bug where output was not printed until a newline was sent from the device using non-defmt-ed RTT
+
+[#77]: https://github.com/knurling-rs/probe-run/pull/77
+
+### Changed
+
+- [#25] increased RTT attach retries. Needed, in some cases, for inter-operation with `rtt-target`
```
jonas-schievink, line 76:
```suggestion
- [#25] increased RTT attach retries, which is sometimes needed for inter-operation with `rtt-target`
```

jonas-schievink (commented):
Did we make any breaking changes to the CLI that would require making this 0.2.0?

Also, do we want to wait until the defmt release to include defmt support in this release? That would save us from doing the weird editing steps before publishing.

japaric:
> Did we make any breaking changes to the CLI that would require making this 0.2.0?

I don't think so but we did change the condition for non-zero exit code (HardFault = non-zero exit code; regardless of whether panic-probe is used or not) which also changed the behavior around hard faults.

> Also, do we want to wait until the defmt release to include defmt support in this release? 

I'd say release next probe-run after defmt is published and have that probe-run release ship with defmt always enabled -- that is we remove the "defmt" Cargo feature, which won't be a semver breaking change because the v0.1.x releases never had a "defmt" Cargo feature

CHANGELOG.md
japaric: I wrote them like this because it make it easier to not miss adding the `[#12]: https` line. Unfortunately, it seems the link doesn't render properly if there's no newline between the bullet and the outline link. I can cluster the links again.

Mirabellensaft (approved):

PR #96, opened 2020-11-11
hard-enable the defmt Cargo feature

japaric:
but keep it around so `cargo install --git $URL --features defmt` continues to work
with this, probe-run will always include defmt support (behind the --defmt command line flag)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ license = "MIT OR Apache-2.0"
 name = "probe-run"
 readme = "README.md"
 repository = "https://github.com/knurling-rs/probe-run"
+# TODO(on 0.2.0) remove the "defmt" Cargo feature
 version = "0.1.3"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -28,7 +28,8 @@ signal-hook = "0.1.16"
 structopt = "0.3.15"
 
 [features]
-defmt = ["defmt-elf2table", "defmt-decoder"]
+# this feature does nothing and it's kept for backwards compatibility; it will be removed in the next semver bump
```
Lotterleben, line 31: nit: would it make sense to note this down as a low-prio issue as a reminder or is that overkill?

japaric: hmm, actually I think adding a TODO comment to this Cargo.toml just above the `version = ` field would be more visible. (That way whoever bumps the minor version next will be reminded to remove this feature)

Lotterleben: ha! yeah that makes sense and causes less clutter

Lotterleben (approved):
Approving to get the PR out of your way but I'm not 100% confident I understand all the implications yet

PR #106, opened 2020-11-19
report program size

### LINKED ISSUE #99: Output size info when flashing
The flashing step currently just outputs "flashing program", it would be nice if it contained a bit more info pertaining to the program it is flashing.

In particular, total flash usage is a small but important piece of info – it lets you estimate how long the flashing step might take, and is a useful metric to keep in mind while developing in general.

japaric:
this reports the size of the part of the program that will loaded to Flash

the count does not include the `.uninit` section (which is marked NOLOAD in cortex-m-rt)

due to a bug in cortex-m-rt (which has already been fixed but the fix is not on crates.io) this will
include `.bss` in the count. the bug also causes probe-rs to load a bunch of zeros, as many as the
size of .bss, to Flash so in a sense the reported value is correct in that it reflects the amount of
work probe-rs will do

fixes #99 

Example output:
``` console
$ cargo run --bin hello
  (HOST) INFO  flashing program (7.64 KiBi)
  (HOST) INFO  success!
0.000000 INFO  Hello, world!
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,7 +27,7 @@ use gimli::{
 };
 use object::{
     read::{File as ElfFile, Object as _, ObjectSection as _},
-    SymbolSection,
+    ObjectSegment, SymbolSection,
 };
 use probe_rs::config::{registry, MemoryRegion, RamRegion};
 use probe_rs::{
@@ -313,7 +313,8 @@ fn notmain() -> Result<i32, anyhow::Error> {
         log::info!("skipped flashing");
     } else {
         // program lives in Flash
-        log::info!("flashing program");
+        let size = program_size_of(&elf);
+        log::info!("flashing program ({:.02} KiBi)", size as f64 / 1024 as f64);
         flashing::download_file(&mut sess, elf_path, Format::Elf)?;
         log::info!("success!");
     }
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -313,7 +313,8 @@ fn notmain() -> Result<i32, anyhow::Error> {
         log::info!("skipped flashing");
     } else {
         // program lives in Flash
-        log::info!("flashing program");
+        let size = program_size_of(&elf);
+        log::info!("flashing program ({:.02} KiBi)", size as f64 / 1024 as f64);
```
jamesmunns, line 317:
```suggestion
        log::info!("flashing program ({:.02} KiB)", size as f64 / 1024 as f64);
```

src/main.rs
```diff
@@ -528,6 +529,11 @@ fn notmain() -> Result<i32, anyhow::Error> {
     )
 }
 
+fn program_size_of(file: &ElfFile) -> u64 {
+    // `segments` iterates only over *loadable* segments, which are the segments that will be loaded to Flash by probe-rs
+    file.segments().map(|segment| segment.size()).sum()
```
jamesmunns, line 534: Stylistic nit: could just `.map(Segment::size)`

jamesmunns (approved):
Small (optional) nits, overall looks good!

src/main.rs
japaric: (did not compile because `segments` yields `Segment` values but `Segment::size` expects a reference (`&Segment`))

PR #109, opened 2020-11-20
don't print <exception entry> more thn once per frame

### LINKED ISSUE #59: `<exception entry>` printed twice on UDF
STR
``` rust
#[cortex_m_rt::entry]
fn main() -> ! {
    cortex_m::asm::udf();
}
```

``` text
$ cargo rb abort
stack backtrace:
   0: 0x000003e0 - HardFaultTrampoline
      <exception entry>
      <exception entry>
   1: 0x00000140 - __udf
   2: 0x00000118 - cortex_m::asm::udf
   3: 0x0000012c - abort::__cortex_m_rt_main
   4: 0x00000122 - main
   5: 0x000000fa - Reset
```

japaric:
as far as I could test `HardFaultTrampoline` always has a special LR value so we always enter the block at line 797 and print `<exception entry>`

fixes #59

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -747,7 +747,6 @@ fn backtrace(
         if top_exception.is_none() {
             top_exception = Some(if pc & !THUMB_BIT == vector_table.hard_fault & !THUMB_BIT {
                 // HardFaultTrampoline
-                println!("      <exception entry>");
 
                 let stack_overflow = if let Some(sp_ram_region) = sp_ram_region {
                     // NOTE stack is full descending; meaning the stack pointer can be `ORIGIN(RAM) +


src/main.rs
```diff
@@ -747,7 +747,6 @@ fn backtrace(
         if top_exception.is_none() {
```
japaric, line 747: I'm not sure the code above this line applies anymore. the code inside the if checks for overflow but doesn't do special logic around unwinding the frame. The frame is unwound normally using the logic in line 797.

probably the code around this can be simplified now that we put a breakpoint on HardFaultTrampoline

jonas-schievink: Yeah, seems like a good idea. I do find the code in probe-run to be a bit hard to follow sometimes.

jonas-schievink (approved):

PR #110, opened 2020-11-23
colorize defmt::assert_eq output

japaric:
this implementation parses a formatted log message and checks if the output corresponds to the `defmt::assert_eq!` macro. If that's the case then it applies color using the same approach as the `pretty-assertions` crate.

a better implementation would be to tag the log messages from the `defmt::assert_eq` and apply the coloring based on the tag, instead of re-parsing the formatted log message. however, that doesn't seem possible to implement in the v0.1.x line because:
- if we add a new variant to `decoder::Tag` that would be a breaking change (UB?) because the enum is not `repr(C)`
- if we try to apply to formatting when the `SymbolTag::Custom` tag is present then old versiosn of probe-run will not print the `assert_eq` message (as they ignore custom tags)

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -164,6 +164,9 @@ fn color_diff(text: String) -> String {
             && right.starts_with(RIGHT_START)
             && right.ends_with(END)
         {
+            const DARK_RED: Colour = Colour::Fixed(52);
+            const DARK_GREEN: Colour = Colour::Fixed(22);
+
             // `defmt::assert_eq!` output
             let left = &left[LEFT_START.len()..left.len() - END.len()];
             let right = &right[RIGHT_START.len()..right.len() - END.len()];
@@ -189,7 +192,7 @@ fn color_diff(text: String) -> String {
                     }
                     Difference::Add(_) => continue,
                     Difference::Rem(s) => {
-                        write!(buf, "{}", Colour::Red.on(Colour::Fixed(52)).bold().paint(s)).ok();
+                        write!(buf, "{}", Colour::Red.on(DARK_RED).bold().paint(s)).ok();
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


jonas-schievink:
> if we add a new variant to `decoder::Tag` that would be a breaking change (UB?) because the enum is not `repr(C)`

Did you mean `#[non_exhaustive]` instead of `#[repr(C)]` here? (I agree with the conclusion anyways)

src/logger.rs
```diff
@@ -142,3 +145,77 @@ impl Log for Logger {
...
+            .ok();
+            write!(buf, "{}", "<".red()).ok();
+            for diff in &changeset.diffs {
+                match diff {
+                    Difference::Same(s) => {
+                        write!(buf, "{}", s.red()).ok();
+                    }
+                    Difference::Add(_) => continue,
+                    Difference::Rem(s) => {
+                        write!(buf, "{}", Colour::Red.on(Colour::Fixed(52)).bold().paint(s)).ok();
```
jonas-schievink, line 192: Can you move the 52 to a constant?

src/logger.rs
```diff
@@ -142,3 +145,77 @@ impl Log for Logger {
...
+                match diff {
+                    Difference::Same(s) => {
+                        write!(buf, "{}", s.green()).ok();
+                    }
+                    Difference::Rem(_) => continue,
+                    Difference::Add(s) => {
+                        write!(
+                            buf,
+                            "{}",
+                            Colour::Green.on(Colour::Fixed(22)).bold().paint(s)
```
jonas-schievink, line 209: (same here)

jonas-schievink (commented):
Cool!

japaric:
> Did you mean #[non_exhaustive] instead of #[repr(C)] here? (I agree with the conclusion anyways)

yes. (in my mind, repr(C) (+ lots of care) is needed if you serialize C-like enums that you'd like to extend without breaking things. we are not binary (de)serializing the tags so non_exhaustive should suffice)

PR #117, opened 2020-12-02
Wait for breakpoint to actually hit before switching RTT to blocking.

Dirbaio:
Without this I was seeing my firmware boot without RTT being changed to blocking mode. 

I guess probe-run would change mode before the RTT control block is actually initialized, so it'd get overwritten.


japaric (approved):
Whoops; I forgot to add this. Nice catch; thank you!

PR #121, opened 2020-12-15
add bors

japaric:


japaric:
bors ping

.github/workflows/ci.yml
```diff
@@ -37,3 +38,25 @@ jobs:
...
+
+  ci-failure:
+    name: ci
+    if: ${{ !success() }}
+    needs:
+      - test
+    runs-on: ubuntu-20.04
+    steps:
+      - name: CI failed
+        run: exit 1
```
jonas-schievink, line 62: Hmm, why are these needed? Can't you add all required workflows to bors.toml instead? `cortex-m` doesn't have this, for example.

bors.toml
```diff
@@ -0,0 +1,2 @@
+delete_merged_branches = true
+status = ["ci"]
```
jonas-schievink, line 2: This file can be moved to `.github/bors.toml` to keep the root dir clean

.github/workflows/ci.yml
japaric: unsure; copied them from a working GHA+bors configuration (heapless)

PR #126, opened 2020-12-21
Print probe list when multiple probes were found

### LINKED ISSUE #118: Print probe list when multiple probes were found
When probe-run errors because it finds multiple probes, it would be helpful to print the probe list, so that the user can easily specify it without having to run `probe-run --list-probes`

hasheddan:
Prints all discovered probes when error is multiple probes found. Adapts
print_probes to accept a Vec such that it does not list probes after
they were already listed previously.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Fixes #118

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -116,7 +116,7 @@ fn notmain() -> Result<i32, anyhow::Error> {
     logger::init(opts.verbose);
 
     if opts.list_probes {
-        return print_probes();
+        return print_probes(Probe::list_all());
     }
 
     if opts.list_chips {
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Nice, thanks!

Can you post an example output of this?

hasheddan:
@jonas-schievink sure thing!
```
The following devices were found:
0: <probe-rs probe 1 debug output>
1: <probe-rs probe 2 debug output>
Error: more than one probe found; use --probe to specify which one to use
```

jonas-schievink:
Looks good, thanks!

bors r+

PR #105, opened 2020-11-16
Autodetect defmt usage

jonas-schievink:
Closes #102

PR #129, opened 2021-01-06
reject use of defmt and --no-flash (again)

japaric:
before PR #105 using defmt required the `--defmt` flag. At that time it was not possible to use the `--defmt` and `--no-flash` flags together. The rationale for that restriction is that without it `cargo run` can invoke `probe-run` with a newer version of the ELF different from the ELF that was last flashed on the device and that can result in log messages that make no sense and/or decoder errors.

After PR 105 landed it unintentionally became possible to use defmt with --no-flash due to the auto-detection (auto-enable) mechanism. This PR corrects that and makes the use of `--no-flash` flag and defmt an error again.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -385,9 +385,6 @@ fn notmain() -> Result<i32, anyhow::Error> {
         core.run()?;
     }
 
-    // Print a separator before the device messages start.
-    eprintln!("{}", "─".repeat(80).dimmed());
-
     let exit = Arc::new(AtomicBool::new(false));
     let sig_id = signal_hook::flag::register(signal_hook::SIGINT, exit.clone())?;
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -399,6 +396,10 @@ fn notmain() -> Result<i32, anyhow::Error> {
         .as_ref()
         .map_or(false, |ch| ch.name() == Some("defmt"));
 
+    if use_defmt && opts.no_flash {
+        bail!("\"defmt\" RTT channel is in use, but `--no-flash` was specified -- this combination is not allowed");
```
Lotterleben, line 400: nit: `"defmt" RTT channel is in use` requires system knowledge in order to grasp what the problem is– could this be rephrased to something like "attempting to specify `--no-flash`while using `defmt` -- this combination is not allowed"?

Lotterleben (approved):

PR #130, opened 2021-01-12
add defmt-print

japaric:
this adds a tool called `defmt-print` that decodes a defmt stream and prints the formatted output to the console. The new tool uses the same format as `probe-run`
this can be used to decode defmt data send over serial port, ITM/SWO, etc.
``` console
$ # itm.txt produced by OpenOCD
$ itmdump -F -f itm.txt | defmt-print -e my-elf

$ # UART output
$ cat /dev/ttyUSB0 | defmt-print -e my-elf
```

### DIFF
--- a/src/bin/defmt-print.rs
+++ b/src/bin/defmt-print.rs
@@ -1,0 +1,55 @@
+use std::{
+    env, fs,
+    io::{self, Read},
+    path::PathBuf,
+};
+
+use anyhow::anyhow;
+use structopt::StructOpt;
+
+use probe_run::logger;
+
+/// Prints defmt-encoded logs to stdout
+#[derive(StructOpt)]
+struct Opts {
+    #[structopt(short, parse(from_os_str))]
+    elf: PathBuf,
+    // may want to add this later
+    // #[structopt(short, long)]
+    // verbose: bool,
+    // TODO add file path argument; always use stdin for now
+}
+
+const READ_BUFFER_SIZE: usize = 1024;
+
+fn main() -> anyhow::Result<()> {
+    let opts: Opts = Opts::from_args();
+    let verbose = false;
+    logger::init(verbose);
+
+    let bytes = fs::read(&opts.elf)?;
+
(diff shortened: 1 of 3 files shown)


jonas-schievink (approved):
Changes look good, though I'd expect a tool like this to be in the defmt repo, not probe-run. I guess that fits into the longer-term refactoring and cleanup tasks though.

src/bin/defmt-print.rs
```diff
@@ -0,0 +1,55 @@
...
+use anyhow::anyhow;
+use structopt::StructOpt;
+
+use probe_run::logger;
+
+/// Prints defmt-encoded logs to stdout
+#[derive(StructOpt)]
+struct Opts {
+    #[structopt(short, parse(from_os_str))]
+    elf: PathBuf,
```
jonas-schievink, line 16: Is there a reason this is `-e` instead of a positional argument?

japaric:
yeah, it's a bit unfortunate that `cargo install probe-run` install two binaries but having `defmt-print` in the defmt would either require duplicating code or refactoring probe-run's stdout format into an unstable library (which wouldn't have that much use?)

src/bin/defmt-print.rs
japaric: Using a flag here would let us add an *optional* positional argument that specifies the defmt file to decode. That is these two would work:
``` console
$ cat /dev/ttyUSB0 | defmt-print -e my-elf
$ defmt-print -e my-elf some-defmt-data.bin
```

japaric: > add an optional positional argument

add backwards compatibly in a patch release

jonas-schievink:
I'd be fine with adding the defmt -> log adapter code to the defmt repo. That seems potentially useful for any defmt printer.

src/bin/defmt-print.rs
jonas-schievink: Ah, I see. I suppose we could also make the input file a named argument, but this works too.

japaric:
closing in favor of knurling-rs/defmt#338

PR #133, opened 2021-01-17
Do not display full version with --help

### LINKED ISSUE #131: --help output should probably not include "supported defmt version"
**Describe the bug**

`--help` output includes "supported defmt version: .."; that bit of output should only be printed when `--version` is used

**To Reproduce**

Install git version then do `probe-run --help`.

Got:
``` console
$ probe-run --help
probe-run 0.1.8 (f59c74a 2021-01-07)
supported defmt version: c4461eb148461e54d9a1e203cf01393f07ce84f0
A Cargo runner for microcontrollers
(..)
```

Expected:
``` console
$ probe-run --help
probe-run 0.1.8
A Cargo runner for microcontrollers
(..)
```

Other comments: `probe-run --version` output looks OK

Javier-varez:
Instead of using App::version from clap use a custom version flag that doesn't change the output of --help or -h.

Closes #131

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -44,28 +44,22 @@ fn main() -> Result<(), anyhow::Error> {
 }
 
 // the string reported by the `--version` flag
-fn version() -> &'static str {
-    // version from Cargo.toml e.g. "0.1.4"
-    let mut output = env!("CARGO_PKG_VERSION").to_string();
+fn print_version() -> Result<i32, anyhow::Error> {
+    println!(
+        "{}{}",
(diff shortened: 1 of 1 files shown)


japaric (approved):
Nice! Thanks for the fix

japaric:
( we should do the same in `defmt-print` https://github.com/knurling-rs/defmt/blob/574f77f05f0d72f905c8955cb7a652fb4f91726e/print/src/main.rs#L95 )

PR #141, opened 2021-02-04
Satisfy clippy

Urhengulas:


src/main.rs
```diff
@@ -323,7 +316,7 @@ fn notmain() -> Result<i32, anyhow::Error> {
     } else {
         // program lives in Flash
         let size = program_size_of(&elf);
-        log::info!("flashing program ({:.02} KiB)", size as f64 / 1024 as f64);
+        log::info!("flashing program ({:.02} KiB)", size as f64 / 1024_f64);
```
jonas-schievink, line 319: Even simpler:

```suggestion
        log::info!("flashing program ({:.02} KiB)", size as f64 / 1024.0);
```

src/main.rs
```diff
@@ -680,6 +673,7 @@ impl<'c, 'probe> Registers<'c, 'probe> {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
jonas-schievink, line 676: This is something we should fix at some point, the code in probe-run has become pretty messy

```suggestion
#[allow(clippy::too_many_arguments)] // FIXME: clean this up
```

Urhengulas: Is this type of comment automatically recognized by some tool?

Urhengulas: Or should we maybe just open an issue for it, since this is more visible?

Urhengulas:
Thanks @jonas-schievink

PR #151, opened 2021-02-15
discard arguments that come after the ELF file path

japaric:
including unknown flags
fixes knurling-rs/defmt#260

which results in `probe-run` being invoked like this:
`probe-run --chip CHIP some-elf module::path --exact --capture`
the last 3 flags are now accepted by the parser but then discarded


jonas-schievink (approved):
Seems like a good workaround!

bors r+

PR #153, opened 2021-02-22
Update to 0.2.0

jonas-schievink:


Cargo.toml
```diff
@@ -29,7 +28,3 @@ rustc-demangle = "0.1.16"
 signal-hook = "0.3.4"
 structopt = "0.3.15"
 hidapi = "1.2.5"
-
-[features]
-# this feature does nothing and it's kept for backwards compatibility; it will be removed in the next semver bump
-defmt = []
```
japaric, line 35: we should also remove the `defmt` field form the `Opts` field

japaric (changes requested):
requested a change; the rest looks good to me

Cargo.toml
jonas-schievink: Ah, right

japaric (approved):
:+1:

PR #161, opened 2021-02-24
remind the user that the `bench` profile should be also overridden

japaric:
with `debug = 1` to avoid running into #9


jonas-schievink (approved):
bors r+

PR #162, opened 2021-02-24
Remove panic-probe

jonas-schievink:
It now lives in the defmt repo


japaric (approved):
I agree. Enough time has passed since people were using git dependencies that pointed to this repo so it should be safe to remove.

Issue #138, opened 2021-01-28
report exit reason, optional backtrace


japaric:
`asm::bkpt` halts the microcontroller; when probe-run sees this it exits with exit-code = 0 and also prints a stack backtrace.

For this scenario, I think the backtrace should be optional to match the behavior of `std` program (`process::exit(0)` *never* prints a backtrace).

It would also be nice if `probe-run` reported *after* the target's (defmt's) logs that it exited *because* the target device halted. Likewise, if the device panicked probe-run should say that the exit reason was a panic. Ditto for stack overflow.

I'm thinking of something like this:

``` rust
fn main() -> ! {
    defmt::info("Hello, world");
    loop { asm::bkpt() }
}
```

``` console
$ cargo run --bin hello
  (HOST) INFO  flashing program (19.79 KiB)
  (HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
0.000000 INFO  Hello, world!
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:9
────────────────────────────────────────────────────────────────────────────────
  (HOST) INFO  device halted; exiting. To see the backtrace at the exit point set the `PROBE_RUN_BACKTRACE` env var
```

``` console
$ PROBE_RUN_BACKTRACE=1 cargo run --bin hello
  (HOST) INFO  flashing program (19.79 KiB)
  (HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
0.000000 INFO  Hello, world!
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:9
────────────────────────────────────────────────────────────────────────────────
  (HOST) INFO  device halted; exiting.
stack backtrace:
   0: __bkpt
   1: hello::__cortex_m_rt_main
        at src/bin/hello.rs:11
   2: main
        at src/bin/hello.rs:7
   3: ResetTrampoline
        at /home/japaric/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-rt-0.6.13/src/lib.rs:547
   4: Reset
        at /home/japaric/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-rt-0.6.13/src/lib.rs:550
```

I would also print an horizontal bar after the device logs and put the stack backtrace *after* that. The stuff that comes after that second horizontal bar is stuff that happens on the host (the backtrace is computed on the host).

Lotterleben:
[...]

Issue #127, opened 2020-12-29
backtrace can infinte-loop.


Dirbaio:
I'm seeing this behavior with `-C force-frame-pointers=no`. 

I think it's to be expected that backtracing doesn't work correctly with it, but I think at least this should be detected and fail with `error: the stack appears to be corrupted beyond this point` instead of looping forever.

If there's interest I can try cooking a binary that reproduces this.

```
stack backtrace:
   0: HardFaultTrampoline
      <exception entry>
   1: tester_gwc::sys::__cortex_m_rt_WDT
        at ak/src/bin/../sys.rs:556
   2: WDT
        at ak/src/bin/../sys.rs:553
      <exception entry>
   3: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
        at /home/dirbaio/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.8/src/future/select.rs:95
   4: tester_gwc::common::abort_on_keypress::{{closure}}
        at ak/src/bin/../tester_common.rs:26
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
        at /home/dirbaio/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80
   6: tester_gwc::test_network::{{closure}}
        at ak/src/bin/tester_gwc.rs:61
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
        at /home/dirbaio/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80
   8: tester_gwc::main::{{closure}}
        at ak/src/bin/tester_gwc.rs:43
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  10: tester_gwc::sys::main_task::task::{{closure}}
        at ak/src/bin/../sys.rs:196
  11: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
        at /home/dirbaio/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80
  12: embassy::executor::Task<F>::poll
        at /home/dirbaio/akiles/embassy/embassy/src/executor/mod.rs:132
  13: core::cell::Cell<T>::get
[...]

PR #163, opened 2021-02-26
Report exit reason, make backtrace optional

Lotterleben:
fixes https://github.com/knurling-rs/probe-run/issues/138

I've also tried to improve readability as a drive-by by adding some comments and e.g. making `construct_backtrace()`return `Ok(None)` in case of no exception instead of `Ok(TopException::Other)`

Note: I'm not quite sure I caught all the exit reasons or split them up right.

I've tested this PR locally with `panic.rs`, `hello.rs` and `stack_overflow.rs` from our [embedded trainings](https://github.com/ferrous-systems/embedded-trainings-2020).

logs look like this:

## hello

1. without forcing backtrace
2. with forcing backtrace
3. with forced backtrace and `--verbose`

Screenshot 2021-02-25 at 18 26 36

Screenshot 2021-02-26 at 16 18 14
## panic

1. without forcing backtrace
2. with forcing backtrace using ` PROBE_RUN_BACKTRACE=1` (cut off because it's the same)
3. with `--verbose`

Screenshot 2021-02-25 at 18 27 57
)

Screenshot 2021-02-26 at 16 16 59

## stack_overflow

1. without forcing backtrace
2. with forcing backtrace using ` PROBE_RUN_BACKTRACE=1` (cut off because it's the same)

Screenshot 2021-02-26 at 16 15 48


src/main.rs
```diff
@@ -107,9 +107,13 @@ struct Opts {
     /// Arguments passed after the ELF file path are discarded
     #[structopt(name = "REST")]
     _rest: Vec<String>,
+
+    #[structopt(long, env = "PROBE_RUN_BACKTRACE")]
+    force_backtrace: Option<String>,
 }
 
 fn notmain() -> Result<i32, anyhow::Error> {
+
```
Urhengulas, line 116:
```suggestion
```

src/main.rs
```diff
@@ -107,9 +107,13 @@ struct Opts {
     /// Arguments passed after the ELF file path are discarded
     #[structopt(name = "REST")]
     _rest: Vec<String>,
+
+    #[structopt(long, env = "PROBE_RUN_BACKTRACE")]
+    force_backtrace: Option<String>,
```
Urhengulas, line 112: Move before `_rest`? And add doc-comment?

src/main.rs
```diff
@@ -539,30 +545,53 @@ fn notmain() -> Result<i32, anyhow::Error> {
...
 
     Ok(
-        if let Some(TopException::HardFault { stack_overflow }) = top_exception {
-            if stack_overflow {
+        match top_exception {
+            Some(TopException::StackOverflow) => {
                 log::error!("the program has overflowed its stack");
+                SIGABRT
             }
+            Some(TopException::HardFault) => {
```
Urhengulas, line 577: Maybe name `TopException::Panic`, if it is indicating a panic?

src/main.rs
```diff
@@ -539,30 +545,53 @@ fn notmain() -> Result<i32, anyhow::Error> {
...
+                log::error!("the program panicked");
+                SIGABRT
+            }
+            None => {
+                if !force_backtrace {
+                    let mut re_run_args = "PROBE_RUN_BACKTRACE=1 probe-run".to_string();
+                    for arg in std::env::args().skip(1) {
+                        re_run_args = format!("{} {}", re_run_args, arg);
+                    }
```
Urhengulas, line 586: This way of constructing the string seems a bit hacky. Is there maybe a more elegant iterator-based one?

src/main.rs
```diff
@@ -701,6 +730,7 @@ fn backtrace(
     sp_ram_region: &Option<RamRegion>,
     live_functions: &HashSet<&str>,
     current_dir: &Path,
+    force_backtrace: bool
```
Urhengulas, line 733: Clippy already told us that we are using too many arguments 🙈 But this should probably be tackled separately.

src/main.rs
```diff
@@ -737,6 +768,48 @@ fn backtrace(
...
+                // lr`) is executed so special handling is required
+                // also note that hard fault will always be the first frame we unwind
+
+                print_backtrace_start();
+
+                let mut stack_overflow = false;
+
+                if let Some(sp_ram_region) = sp_ram_region {
+                    // NOTE stack is full descending; meaning the stack pointer can be `ORIGIN(RAM) +
+                    // LENGTH(RAM)`
```
Urhengulas, line 786:
```suggestion
                    // NOTE stack is full descending; meaning the stack pointer can be `ORIGIN(RAM) + LENGTH(RAM)`
```

src/main.rs
```diff
@@ -746,7 +819,7 @@ fn backtrace(
                     .transpose()?
                     .unwrap_or(Cow::Borrowed("???"));
 
-                println!("{:>4}: {}", frame_index, name);
+                backtrace_display_str = format!("{}{:>4}: {}\n", backtrace_display_str, frame_index, name);
```
Urhengulas, line 822: What do you think of doing this? This makes it faster clear that we are extending the current backtrace for me.
```suggestion
                backtrace_display_str.push_str(format!("{:>4}: {}\n", frame_index, name));
```

src/main.rs
```diff
@@ -819,20 +872,41 @@ fn backtrace(
...
 
         let lr = registers.get(LR)?;
-        log::debug!("lr=0x{:08x} pc=0x{:08x}", lr, pc);
+
         if lr == LR_END {
             break;
         }
 
+        let exception_entry = lr > 0xffff_ffe0;
+        let stack_corrupted = !cfa_changed && lr & !THUMB_BIT == pc & !THUMB_BIT;
```
Urhengulas, line 881: Is it possible to clarify what this line expresses with a comment?

src/main.rs
```diff
@@ -819,20 +872,41 @@ fn backtrace(
...
+
         // If the frame didn't move, and the program counter didn't change, bail out (otherwise we
         // might print the same frame over and over).
         // Since we strip the thumb bit from `pc`, ignore it in this comparison.
-        if !cfa_changed && lr & !THUMB_BIT == pc & !THUMB_BIT {
+        if stack_corrupted {
             println!("error: the stack appears to be corrupted beyond this point");
-            return Ok(top_exception);
+
+            if let Some(TopException::StackOverflow) = top_exception {
```
Urhengulas, line 902:
```suggestion
            if top_exception == Some(TopException::StackOverflow) {
```

src/main.rs
```diff
@@ -936,6 +1010,15 @@ fn print_chips() -> Result<i32, anyhow::Error> {
     Ok(0)
 }
 
+/// Print a separator before the device messages start.
```
Urhengulas, line 1013: I guess this comment needs to be updated

src/main.rs
```diff
@@ -936,6 +1010,15 @@ fn print_chips() -> Result<i32, anyhow::Error> {
...
+fn print_separator() {
+    eprintln!("{}", "─".repeat(80).dimmed());
+}
+
+fn print_backtrace_start() {
+    println!("{}", "stack backtrace:".dimmed());
```
Urhengulas, line 1019: Should this also be `eprintln!`? Or should we use `println!` 4 lines above?

Urhengulas (changes requested):

src/main.rs
Urhengulas: As mentioned later we can maybe use `String::push_str`?

src/main.rs
Urhengulas: There are some similar lines below

src/main.rs
Lotterleben: yeah :D I think there's a lot of rom for refactoring here, tried to make up for this sind by improving some others :)

src/main.rs
Lotterleben: that would make the line longer than 100 chars (since I almost always look at 2 files simultaneously in vertical split view I tend to be very strict about this :D). I've changed the comment so that `ORIGIN(RAM) + LENGTH(RAM)` is all on one line though

src/main.rs
Lotterleben: ahhh perfect that was the non-hacky solution I was looking for, thank you

src/main.rs
Lotterleben: good catch. I think there's no need to print the line to stderr :D

src/main.rs
Lotterleben: Hrrrm not sure– it's only indicating a panic because we know that `panic-probe` will trigger a HardFault on panic, no? I'd rather not bake this assumption further into our code– deducting "this was a panic" from "some HardFault occurred" is already muddying the waters imo so I'd like to contain this to this questionable match arm :D

src/main.rs
Lotterleben: Do you mean something like `// check for conditions under which we definitely need to print a backtrace` ?

src/main.rs
```diff
@@ -104,6 +104,10 @@ struct Opts {
     #[structopt(short = "V", long)]
     version: bool,
 
+    /// Print a backtrace even if the program ran successfully
+    #[structopt(long, env = "PROBE_RUN_BACKTRACE")]
+    force_backtrace: Option<String>,
```
jonas-schievink, line 109: Is it useful to pass `--force-backtrace=false`? If not, using this would be simpler (you'd pass `--force-backtrace` to enable it):

```suggestion
    force_backtrace: bool,
```

src/main.rs
```diff
@@ -737,6 +770,48 @@ fn backtrace(
             false
         };
 
+       // This is our first run through the loop, some initial handling and printing is required
```
jonas-schievink, line 773:
```suggestion
        // This is our first run through the loop, some initial handling and printing is required
```

src/main.rs
jonas-schievink: You can move the comment from below to the `stack_corrupted` variable:

```rust
// If the frame didn't move, and the program counter didn't change, bail out (otherwise we
// might print the same frame over and over).
// Since we strip the thumb bit from `pc`, ignore it in this comparison.
```

Also this might be insufficient and cause https://github.com/knurling-rs/probe-run/issues/127, so if you'd like to look into that later, this condition would be where I'd start.

src/main.rs
Lotterleben: much nicer! done

src/main.rs
```diff
@@ -737,6 +766,48 @@ fn backtrace(
             false
         };
 
+        // This is our first run through the loop, some initial handling and printing is required
+        // TODO refactor this
+        if frame_index == 0 {
+            if pc & !THUMB_BIT == vector_table.hard_fault & !THUMB_BIT {
```
spookyvision, line 772: that's a pretty deep & long nesting, I'd extract into function

src/main.rs
```diff
@@ -737,6 +766,48 @@ fn backtrace(
...
+                let mut stack_overflow = false;
+
+                if let Some(sp_ram_region) = sp_ram_region {
+                    // NOTE stack is full descending; meaning the stack pointer can be
+                    // `ORIGIN(RAM) + LENGTH(RAM)`
+                    let range = sp_ram_region.range.start..=sp_ram_region.range.end;
+                    stack_overflow = !range.contains(&sp);
+
+                    // if a stack overflow happened, we're definitely printing a backtrace
+                    print_backtrace = print_backtrace || stack_overflow;
```
spookyvision, line 789: could do `print_backtrace |= stack_overflow;` but that's a matter of taste

src/main.rs
```diff
@@ -775,37 +846,19 @@ fn backtrace(
...
-                };
-
-                TopException::HardFault { stack_overflow }
-            } else {
-                TopException::Other
-            });
+        if print_backtrace {
+            // we need to print everything we've collected up until now, otherwise the
+            // debug level logs won't match up
+            print!("{}", backtrace_display_str);
```
spookyvision, line 856: personally I'd put the `\n` not inside the str and use `println!("{}", bt_without_newline)`, but *again* a matter of taste

src/main.rs
Lotterleben: totally– that's why I put a // TODO refactor this in there ;) My secret plan was to do this as a drive-by when I tackle the next issue. Didn't want to overhaul everything in one pass because then

my functionality changes would get lost, making it harder to review
time budget is limited

README.md
```diff
@@ -96,6 +96,8 @@ fn main() -> ! {
 }
 ```
 
+TODO dont forget to adjust the examples
```
Lotterleben, line 99: note to self: resolve this 🤦‍♀️

src/main.rs
Lotterleben: in principle I totally agree; the reason I went for for print! here is that `print_backtrace` is somposed differently depending on which branches are taken beforehand so figuring out which line will be the last to be added would make things more complicated in the current state of the code I think

src/main.rs
Lotterleben: agreed and fixed

README.md
```diff
@@ -96,91 +96,79 @@ fn main() -> ! {
...
+    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
+     Running `probe-run --chip nRF52840_xxAA target/thumbv7em-none-eabihf/debug/hello`
+  (HOST) INFO  flashing program (30.22 KiB)
+  (HOST) INFO  success!
+────────────────────────────────────────────────────────────────────────────────
+INFO:hello -- Hello, world!
+────────────────────────────────────────────────────────────────────────────────
+  (HOST) INFO  exiting because the device halted.
+To see the backtrace at the exit point repeat this run with
+`probe-run --chip nRF52840_xxAA target/thumbv7em-none-eabihf/debug/hello --force-backtrace`
```
jonas-schievink, line 112: Do we really want to print this after every run? Seems like a lot of noise to me

Lotterleben: hrmm. I was trying to be helpful but i see your point– there are other config options but we don't print these either. dialed this down to `device halted without error` & squashed

Urhengulas (approved):

PR #165, opened 2021-02-28
Various simplifications

Urhengulas:
* Simplify imports
* Simplify value comparison
* Add `const SIGSUCCESS: i32 = 0`
* Refactor `fn print_version`
* Remove wrapping with `Result<_, anyhow::Error>`
* Refactor `fn print_chips`
* Refactor `fn notmain`
* Change order of functions
*  Move `struct Stacked` into own private module
* Move `struct Registers` into own private module 
*  Use `anyhow::Result`
* Refactor `mod registers` 

Please refer to the individual commit diffs, cause some changes seems a bit weird in the combined PR diff.

---

I'd like to add a doc-comment to `struct Registers`. Can someone please provide one? 😄

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,21 +43,6 @@ const SIGSUCCESS: i32 = 0;
 const THUMB_BIT: u32 = 1;
 const TIMEOUT: Duration = Duration::from_secs(1);
 
-fn main() -> Result<(), anyhow::Error> {
-    notmain().map(|code| process::exit(code))
-}
-
-// the string reported by the `--version` flag
-fn print_version() -> i32 {
-    const VERSION: &str = env!("CARGO_PKG_VERSION"); // version from Cargo.toml e.g. "0.1.4"
-    const HASH: &str = include_str!(concat!(env!("OUT_DIR"), "/git-info.txt")); // "" OR git hash e.g. "34019f8" -- this is generated in build.rs
-    println!(
-        "{}{}\nsupported defmt version: {}",
-        VERSION, HASH, DEFMT_VERSION
-    );
-    SIGSUCCESS
-}
-
 /// A Cargo runner for microcontrollers.
 #[derive(StructOpt)]
 #[structopt(name = "probe-run", setting = AppSettings::TrailingVarArg)]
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -909,7 +826,18 @@ fn probes_filter(probes: &[DebugProbeInfo], selector: &ProbeFilter) -> Vec<Debug
         .collect()
 }
 
-fn print_probes(probes: Vec<DebugProbeInfo>) -> Result<i32, anyhow::Error> {
+fn print_chips() -> i32 {
```
jonas-schievink, line 829: Let's either return an `anyhow::Result<()>` or nothing from these helpers. We should only deal with raw exit codes when necessary, since they're quite non-Rusty.

src/main.rs
```diff
@@ -20,45 +20,36 @@ use addr2line::fallible_iterator::FallibleIterator as _;
...
-const TIMEOUT: Duration = Duration::from_secs(1);
+use crate::{
+    registers::{Registers, LR, LR_END, PC, SP},
+    stacked::Stacked,
+};
+
 const STACK_CANARY: u8 = 0xAA;
+const SIGABRT: i32 = 134;
+/// Successfull termination of process.
+const SIGSUCCESS: i32 = 0;
```
jonas-schievink, line 50: Success is not a signal, so this should not start with `SIG`. libc calls this `EXIT_SUCCESS`.

Urhengulas: Okay. I was looking into linuxs signals for 15 minutes, but didn't really find a proper solution, so I just made up this term 😓

src/main.rs
Urhengulas: And then `match` it in `fn main`?

This makes it more complex, but we could also let `fn notmain` return `anyhow::Result<Option<i32>>` and then just pass on `None` and `process::exit` on `Some(...)`?

jonas-schievink: If these all return `()`, `notmain` can just always return `Ok(0)`

Urhengulas:
@jonas-schievink How would you summarize what `struct Regsiters` does?

jonas-schievink:
It caches and tracks the state of CPU registers while the stack is being unwound

Urhengulas:
> caches and tracks the state of CPU registers while the stack is being unwound

This does for a nice doc-comment. Thanks! 😄 

I think we are good to go now?

PR #171, opened 2021-03-08
Introduce even more verbose log level

### LINKED ISSUE #169: Add a way to see probe-rs logs.
There's no way to see logs from probe-rs.

- `--verbose` doesn't work
- `RUST_LOG=trace` doesn't work.

This is a blocker in helping people coming to the probe-rs matrix channel asking for help with flashing issues etc. There's no way for them to provide the logs we need without patching and recompiling probe-run.

This is due to this check here https://github.com/knurling-rs/probe-run/blob/ca0b3c53e28bf607a9db03c01798778a17a967c8/src/main.rs#L117-L121

Urhengulas:
Fixes #169

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -88,8 +88,8 @@ struct Opts {
     connect_under_reset: bool,
 
     /// Enable more verbose logging.
-    #[structopt(short, long)]
-    verbose: bool,
+    #[structopt(short, long, parse(from_occurrences))]
+    verbose: u32,
 
     /// Prints version information
     #[structopt(short = "V", long)]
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -88,8 +88,8 @@ struct Opts {
     connect_under_reset: bool,
 
     /// Enable more verbose logging.
-    #[structopt(short, long)]
-    verbose: bool,
+    #[structopt(short, long, multiple(true))]
+    _verbose: bool,
```
jonas-schievink, line 92: This will put the number of `-v` flags into the `verbose` field:

```suggestion
    #[structopt(short, long, parse(from_occurrences))]
    verbose: u32,
```

Urhengulas: Oh wow. I've tried that, but I didn't work because I misspelled "occurrences"...

Issue #172, opened 2021-03-09
Building probe-run against specific defmt


ianrrees:
A couple times, I've found it necessary to build my own probe-run (most recently, to use a newer probe-rs), but doing that leads to a defmt version mismatch problem that can be difficult to resolve.  Is there a way to build probe-run against specific version of defmt?

Specifically, if I do `cargo install probe-run`, I currently get probe-run 0.2.1, with supported defmt version 0.2, however doing `cargo install --path .` in probe-run source checked out at v0.2.1 yields a probe-run with supported defmt version 60c6447f8ecbc4ff023378ba6905bcd0de1e679f .

I've tried using Cargo's patch section to update my firmware's defmt version from only one place, but haven't managed to get that working.

Urhengulas:
Hi @ianrrees, Can you please check if #174 is a viable workaround for you?

ianrrees:
Yep, works great, thanks!

PR #174, opened 2021-03-09
Skip defmt version check

Urhengulas:
workaround for #172 

I started out using `fn std::env::var(...)`, which inspects the env var at runtime, but thought that it makes more sense to check it during compile time using `option_env!`, because this prevents end users from using this and not actually fixing the mismatch.

While on it I added some refactoring. 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,14 @@ If they are exposed, you can connect a "stand alone" probe device to the microco
 
 Note that this may involve some soldering if your board does not come with a pre-attached header to plug your debugger into.
 
+### defmt version mismatch
+
+#### end-user
+Follow the instructions in the error message to resolve the mismatch.
+
+#### developer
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -195,6 +195,14 @@ If they are exposed, you can connect a "stand alone" probe device to the microco
 
 Note that this may involve some soldering if your board does not come with a pre-attached header to plug your debugger into.
 
+### defmt version mismatch
+
+#### end-user
+Follow the instructions in the error message to resolve the mismatch.
+
+#### developer
+If you are building `probe-run` from source, you can disable the version check by setting the `PROBE_RUN_IGNORE_VERSION` env variable to `true` or `1` at compile time.
```
spookyvision, line 204: I know it's super nitpicky but I'd change "env" to "environment"

Urhengulas: sure!

PR #181, opened 2021-04-07
README: add copypasteable example how to run from repo

Lotterleben:
This may be trivial but with my bad memory I regularly have to search my zsh history so... might as well put it into the README in case someone else has the same issue 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,19 @@ Follow the instructions in the error message to resolve the mismatch.
 #### developer
 If you are building `probe-run` from source, you can disable the version check by setting the `PROBE_RUN_IGNORE_VERSION` environment variable to `true` or `1` at compile time.
 
+
+### developer: running your locally modified `probe-run`
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,6 +203,19 @@ Follow the instructions in the error message to resolve the mismatch.
 #### developer
 If you are building `probe-run` from source, you can disable the version check by setting the `PROBE_RUN_IGNORE_VERSION` environment variable to `true` or `1` at compile time.
 
+
+### developer: running your locally modified `probe-run`
+
```
Urhengulas, line 208: Maybe we don't need the headline and attach it into the section just before?

```suggestion
```

Urhengulas (changes requested):

README.md
Lotterleben: they serve different purposes though– I just added that particular environment variable because it was part of one of my most recent local probe-run runs 😅

Urhengulas: Okay :)

PR #183, opened 2021-04-20
README.md: Add troubleshooting for use with RTIC

davidlattimore:
Additional context:
https://github.com/knurling-rs/defmt/issues/450

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,37 @@ If they are exposed, you can connect a "stand alone" probe device to the microco
 
 Note that this may involve some soldering if your board does not come with a pre-attached header to plug your debugger into.
 
+### Error: RTT up channel 0 not found
+
+This may instead present as `Error: RTT control block not found in target memory.`
+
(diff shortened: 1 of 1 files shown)


Urhengulas (changes requested):
Generally looks good to me.

One nit: Can you add some more lines to make the code snippets more "copy-pastable"? For the first one it's probably only some import, for the second maybe some type information for `ctx`.

davidlattimore:
I expanded the second one to include the signature of the init function, which includes the ctx argument. `init` and `idle` come from the RTIC procedural macro. I could include a full RTIC program, but that might be a bit too much?

Urhengulas:
> I could include a full RTIC program, but that might be a bit too much?

Yes, that would be too much. Someone facing this problem likely knows the structs, I guess.

Looks good to me now. Thanks a lot!

Urhengulas (approved):
bors r+

PR #184, opened 2021-04-29
add some bounds checking to unwinding

japaric:
probe-rs' memory access API does not perform bounds checking -- I assume because the hardware probe
itself does not do any bounds checking -- so reading non-existing memory results in a value of 0
being returned.

When a program has overflowed its stack, the HardFault exception is called. The exception entry
pushed some registers onto the stack (see Stacked struct). Because there's no or little stack memory
in the stack overflow scenario the registers are pushed outside the RAM region; this results in
register information being lost. Without this information we cannot locate the frame preempted by
HardFault and thus cannot unwind the stack.

Before this PR, the unwinding process didn't perform bounds check when reading the stacked
registers; this resulted in zeroes being reported by probe-rs. The unwinding process continued with
these wrong values, printed junk names as part of the backtrace and ultimately failed.

With this PR, Stacked::read performs a bound check so when the unwinder sees that invalid memory is
trying to be access it stops the unwinding process and reports a warning.

---

New output:
``` console
$ cargo run --bin stack-overflow
(..)
stack backtrace:
   0: HardFaultTrampoline
      <exception entry>
  (HOST) WARN  exception entry pushed registers outside RAM; not possible to unwind the stack
  (HOST) ERROR the program has overflowed its stack
```

Old output:
``` console
stack backtrace:
   0: HardFaultTrampoline
      <exception entry>
   1: {"package":"defmt","tag":"defmt_prim","data":"{=__internal_Display}","disambiguator":"9857217393540545671"}
error: the stack appears to be corrupted beyond this point
  (HOST) ERROR the program has overflowed its stack
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -844,7 +844,18 @@ fn construct_backtrace(
             println!("      <exception entry>");
 
             let sp = registers.get(SP)?;
-            let stacked = Stacked::read(registers.core, sp, fpu)?;
+            let ram_bounds = sp_ram_region
+                .as_ref()
+                .map(|ram_region| ram_region.range.clone())
+                // if no device-specific information, use the range specific in the Cortex-M* Technical Reference Manual
+                .unwrap_or(0x2000_0000..0x4000_0000);
+            let stacked = if let Some(stacked) = Stacked::read(registers.core, sp, fpu, ram_bounds)?
+            {
+                stacked
+            } else {
+                log::warn!("exception entry pushed registers outside RAM; not possible to unwind the stack");
+                return Ok(top_exception);
+            };
 
             registers.insert(LR, stacked.lr);
             // adjust the stack pointer for stacked registers
--- a/src/stacked.rs
+++ b/src/stacked.rs
@@ -1,3 +1,5 @@
+use std::{mem, ops::Range};
+
 use probe_rs::{Core, MemoryInterface};
 
 /// Registers stacked on exception entry.
@@ -14,6 +16,15 @@ pub struct Stacked {
     fpu_regs: Option<StackedFpuRegs>,
 }
 
+fn bounds_check(bounds: Range<u32>, start: u32, len: u32) -> Result<(), ()> {
+    let end = start + len;
+    if bounds.contains(&start) && bounds.contains(&end) {
+        Ok(())
+    } else {
+        Err(())
+    }
+}
+
 impl Stacked {
     /// Number of 32-bit words stacked in a basic frame.
     const WORDS_BASIC: usize = 8;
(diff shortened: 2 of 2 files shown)


src/main.rs
```diff
@@ -844,7 +844,18 @@ fn construct_backtrace(
             println!("      <exception entry>");
 
             let sp = registers.get(SP)?;
-            let stacked = Stacked::read(registers.core, sp, fpu)?;
+            let ram_bounds = sp_ram_region
+                .as_ref()
+                .map(|ram_region| ram_region.range.clone())
+                // if no device-specific information, use the range specific in the Cortex-M* Technical Reference Manual
+                .unwrap_or(0x2000_0000..0x4000_0000);
```
japaric, line 851: although this is what the TRM specifies I know some vendors use `0x1000_0000` for CCRAM (core-coupled RAM)

src/stacked.rs
```diff
@@ -14,23 +16,51 @@ pub struct Stacked {
     fpu_regs: Option<StackedFpuRegs>,
 }
 
+fn bounds_check(bounds: Range<u32>, start: u32, len: u32) -> Result<(), ()> {
```
japaric, line 19: `range.contains(&other_range)` (std API) didn't work so I wrote this

jonas-schievink: Why didn't it work?

src/main.rs
jonas-schievink: That seems like somewhere you might want to put the stack... hmm, can we use the initial value of SP here and round it down to a multiple of `0x1000_0000`?

jonas-schievink (approved):
This looks good so far, added some minor comments

src/main.rs
japaric: I have only ever seen the start of RAM being aligned to `0x1000_0000` but I'm not aware of that being a requirement of the architecture. I do know that VTOR needs to be aligned to a power of 2 that's proportional to the number of vectors in the vector table (learned that one the hard way).

That being said, is there any reason the `sp_ram_region` value can be `None`: that is it's not possible to locate the RAM region the SP belongs to? Perhaps the RAM region data that comes from `probe-rs` is not reliable? Unsure if that info is retrieved from the device itself or hardcoded in `probe-rs`.

src/stacked.rs
japaric: trait bounds not met. `(0..4).contains(&2)` works but `(0..8).contains(&(2..4))` does not compile. I'm not aware of a std API that does "range contains range?"

jonas-schievink: Ah, right

src/main.rs
jonas-schievink: I believe this data comes from a fixed database in probe-rs, which in turn is derived from vendor-provided documentation, so it might not be correct. I haven't seen that happen so far though.

If we do want to be smarter with the fallback here, this isn't really something that needs fixing before this can be merged though.

PR #197, opened 2021-05-12
refactor "print backtrace" code

### LINKED ISSUE #196: max-backtrace-len = {0,1} prints 2 frames
### STR

Run this command from an initialized app-template

``` console
$ cargo b --bin hello
$ probe-run --chip nrf52840 --force-backtrace --max-backtrace-len=0 target/*/debug/hello
(..)
stack backtrace:
   0: lib::inline::__bkpt
        at ./asm/inline.rs:13
   1: __bkpt
        at ./asm/lib.rs:49
  (HOST) WARN  maximum backtrace length of 0 reached; cutting off the rest
               note: re-run with `--max-backtrace-len=<your maximum>` to extend this limit
  (HOST) INFO  device halted without error
```

You get the same output (2 frames) if you set `max-backtrace-len=1`

### Meta

``` console
$ probe-run -V
0.2.2
supported defmt version: 0.2
```

japaric:
this commit refactors the code that prints the backtrace to be comprised of 3 smaller steps

- `unwind::target`, talks to the probe and unwinds the target's program retrieving only a series of PC (program counters). This is fallible due to potential IO errors
- `symbolicate::frames`, turns those raw PC values into function names and locations; it also expands inlined functions. This is infallible. When there's no debuginfo the PC address is used as the function "name".
- `pp::backtrace`, pretty prints the processed frames preserving the current format. this is infallible API-wise (`println!` may still panic however)

there's more work that could have been done but has intentionally not been included in this commit, like making `unwind::target` and/or `symbolicate::frames` lazy to only do as much work as required by the `--max_backtrace_len` flag.

this commit also fixes issue 196 by making `--max_backtrace_len=1` print the right number of frames. The behavior of `--max_backtrace_len=0` has been changed to *never* print a backtrace (`--force-backtrace --max-backtrace-len=0` does *not* print a backtrace)

this commit also (non-exhaustively) groups Cortex-M constants and operations in a module to make them easier to spot (in case we ever add support for other architectures like RISC V). a qualified path, instead of an import, is used to access the items in that module.

we don't have automated tests for the backtrace but manual testing indicates that behavior is preserved. the warning message about the stack being corrupted now has less detail however.

fixes #196

### DIFF
--- a/src/backtrace.rs
+++ b/src/backtrace.rs
@@ -9,8 +9,13 @@ mod pp;
 mod symbolicate;
 mod unwind;
 
+pub(crate) struct Settings<'p> {
+    pub(crate) current_dir: &'p Path,
+    pub(crate) max_backtrace_len: u32,
+    pub(crate) force_backtrace: bool,
+}
+
 /// (virtually) unwinds the target's program and prints its backtrace
-#[allow(clippy::too_many_arguments)]
 pub(crate) fn print(
     core: &mut Core,
     debug_frame: &[u8],
@@ -18,26 +23,29 @@ pub(crate) fn print(
     vector_table: &VectorTable,
     sp_ram_region: &Option<RamRegion>,
     live_functions: &HashSet<&str>,
-    current_dir: &Path,
-    force_backtrace: bool,
-    max_backtrace_len: u32,
+    settings: &Settings,
 ) -> anyhow::Result<Outcome> {
     let unwind = unwind::target(core, debug_frame, vector_table, sp_ram_region)?;
 
-    let frames = symbolicate::frames(&unwind.raw_frames, live_functions, current_dir, elf);
+    let frames = symbolicate::frames(
+        &unwind.raw_frames,
+        live_functions,
+        settings.current_dir,
+        elf,
+    );
 
     let contains_exception = unwind
         .raw_frames
         .iter()
         .any(|raw_frame| raw_frame.is_exception());
 
-    let print_backtrace = force_backtrace
+    let print_backtrace = settings.force_backtrace
         || unwind.outcome == Outcome::StackOverflow
         || unwind.corrupted
         || contains_exception;
 
-    if print_backtrace && max_backtrace_len > 0 {
-        pp::backtrace(&frames, max_backtrace_len);
+    if print_backtrace && settings.max_backtrace_len > 0 {
+        pp::backtrace(&frames, settings.max_backtrace_len);
 
         if unwind.corrupted {
             log::warn!("call stack was corrupted; unwinding could not be completed");
--- a/src/backtrace.rs
+++ b/src/backtrace.rs
@@ -1,0 +1,48 @@
+use std::{collections::HashSet, path::Path};
+
+use object::read::File as ElfFile;
+use probe_rs::{config::RamRegion, Core};
+
+use crate::{Outcome, VectorTable};
+
+mod pp;
+mod symbolicate;
+mod unwind;
+
+/// (virtually) unwinds the target's program and prints its backtrace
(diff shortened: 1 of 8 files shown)


src/backtrace.rs
```diff
@@ -0,0 +1,48 @@
...
+pub(crate) fn print(
+    core: &mut Core,
+    debug_frame: &[u8],
+    elf: &ElfFile,
+    vector_table: &VectorTable,
+    sp_ram_region: &Option<RamRegion>,
+    live_functions: &HashSet<&str>,
+    current_dir: &Path,
+    force_backtrace: bool,
+    max_backtrace_len: u32,
```
jonas-schievink, line 23: I think I agree with Clippy here, it would be nice to group some of these arguments in a struct (maybe have one for backtrace options, one for stuff extracted from the ELF)

src/backtrace/pp.rs
```diff
@@ -0,0 +1,48 @@
...
+                }
+
+                frame_index += 1;
+
+                if frame_index >= max_backtrace_len {
+                    // NOTE whitespace preceding "note: " is intentional and used for alignment
+                    // purposes
+                    log::warn!(
+                        "maximum backtrace length of {} reached; cutting off the rest.
+               note: re-run with `--max-backtrace-len=<your maximum>` to extend this limit",
```
jonas-schievink, line 39: Consider splitting this extra line into its own `log::warn!` invocation, that means we don't have to rely on how many characters the logging sink prepends

src/backtrace/unwind.rs
```diff
@@ -0,0 +1,174 @@
...
+    let mut registers = Registers::new(lr, sp, core);
+    let mut raw_frames = vec![];
+    let mut corrupted = true;
+
+    loop {
+        if cortexm::is_hard_fault(pc, vector_table) {
+            debug_assert!(
+                raw_frames.is_empty(),
+                "when present HardFault handler must be the first frame we unwind but wasn't"
+            );
```
jonas-schievink, line 50: Is this a hard requirement? I'd prefer to make it a regular `assert!` if so. That would mean that HardFaul cannot call any functions though, right?

jonas-schievink (commented):
Looks like a solid improvement, nice!

src/backtrace/unwind.rs
japaric: `probe-run` puts a breakpoint on the HardFault handler (the trampoline in cortex-m-rt, not the user-overridable function) so the program is halted because any function can be called from the exception handler.

jonas-schievink: Ah, right

src/backtrace.rs
japaric: I agree. For now I have grouped the backtrace settings into a struct. That removes 2 arguments and barely passes the clippy lint.

I'll leave the refactor of grouping ELF stuff into its own struct as follow-up work. In my mind, there should a (fallible?) operation that goes from `object::File` into `our::ProcessedElf` which is then passed by reference everywhere. That op should be done early in the program to detect issues like missing debug info early on.

src/backtrace.rs
```diff
@@ -0,0 +1,56 @@
...
+mod unwind;
+
+pub(crate) struct Settings<'p> {
+    pub(crate) current_dir: &'p Path,
+    pub(crate) max_backtrace_len: u32,
+    pub(crate) force_backtrace: bool,
+}
+
+/// (virtually) unwinds the target's program and prints its backtrace
+pub(crate) fn print(
```
Urhengulas, line 19: I know I am late to the party, but why are you using `pub(crate)` although it is a application and not a library?

Issue #139, opened 2021-01-28
colored stack backtrace


japaric:
it would be nice if the stack backtrace was colored, with stack frames that belong to the current crate being highlighted in some way (e.g. bold font)

[color-eyre](https://docs.rs/color-eyre/0.5.10/color_eyre/) does this for example. (color-eyre also hides less relevant frames by default. I personally would like to avoid having to re-run the program with a different env var to see the complete stack backtrace because re-running the program (or making the program panic again) is relatively expensive: you need to re-flash the program, which takes a while.)

Lotterleben:
one die note– I think we should make sure to not overdo it and overwhelm our users with too much color because then the important info gets lost as well.

Maybe we could dial down the 
```
  (HOST) INFO  flashing program (5.96 KiB)
  (HOST) INFO  success!
```
by making that a simple dimmed println instead?

```
    println!("flashing program ({:.02} KiB)".dimmed(), size as f64 / 1024.0);
```

PR #200, opened 2021-05-12
backtrace: highlight frames that point to local code

japaric:
using the heuristic: if the file location of the function is relative to the current directory then
it's "local code"

before:
Screenshot from 2021-05-12 18-47-40
after:
Screenshot from 2021-05-12 18-46-03

I don't have too many opinions about *how* to highlight these frames so treat the above as a strawman proposal

cc #139 @Lotterleben


jonas-schievink:
As a followup, we might be able to check if the path is part of the containing Cargo workspace instead, that might improve the accuracy a bit.

bors r+

Lotterleben:
This is brilliant!

PR #206, opened 2021-05-18
fix unwinding exceptions that push FPU registers onto the stack

japaric:
the correct value of constant EXC_RETURN_MARKER is 0xFFFF_FFE0 but the value was typo-ed in the
refactor part of PR #163 

https://github.com/knurling-rs/probe-run/commit/226c6930dbcd0def0acb2dc51d0584a291c2f2bf#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcL755


jonas-schievink:
Okay no wonder I didn't see this during review :laughing: 

bors r+

PR #207, opened 2021-05-18
unwind: read as little stacked registers as possible

### LINKED ISSUE #187: unwinding: do not read _all_ stacked registers
to unwind pass exceptions we currently read *all* the registers stacked by the exception (up to 25 words):
https://github.com/knurling-rs/probe-run/blob/e8be32a572285e2c004753e525a5ab22707a1006/src/stacked.rs#L5-L15

I believe however that we only need to retrieve the stacked LR and PC registers for the unwinding process so we could reduce the read memory operation to only fetch 7 words. Reading less memory may increase the chances of being able to unwind the stack in presence of a stack overflow (see PR #184 )

japaric:
the other registers that were being read are not used in the unwinding process
closes #187
(this doesn't appear to help with unwinding programs that are in a 'stack overflowed' state however)

### DIFF
--- a/src/stacked.rs
+++ b/src/stacked.rs
@@ -5,11 +5,12 @@ use probe_rs::{Core, MemoryInterface};
 /// Registers stacked on exception entry.
 #[derive(Debug)]
 pub(crate) struct Stacked {
-    r0: u32,
-    r1: u32,
-    r2: u32,
-    r3: u32,
-    r12: u32,
+    // also pushed onto the stack but we don't need to read them
+    // r0: u32,
+    // r1: u32,
+    // r2: u32,
+    // r3: u32,
+    // r12: u32,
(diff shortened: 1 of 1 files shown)


src/stacked.rs
```diff
@@ -4,16 +4,15 @@ use probe_rs::{Core, MemoryInterface};
 
 /// Registers stacked on exception entry.
 #[derive(Debug)]
-pub struct Stacked {
+pub(crate) struct Stacked {
     r0: u32,
     r1: u32,
     r2: u32,
     r3: u32,
     r12: u32,
```
jonas-schievink, line 12: Do we have to keep reading these?

japaric: actually, we don't! new commit avoids reading these first 5 registers

PR #210, opened 2021-05-20
extend `--shorten-paths`; highlight path segments

### LINKED ISSUE #66: backtrace: Shorten paths to crates/libcore
It would be nice to shorten long paths to rust libraries or published dependency crate versions, so your backtrace lines are better human-readable:

For core/alloc/std libraries, a format stating the version of the compiler used in a format similar to `rust-toolchain` file specifications would be good

`[stable-1.46.0]/lib/rustlib/src/foo.rs`

For published library dependencies, it would be good to use a similar format, with `$crate_name-$crate_version:

`[heapless-0.5.2]/src/vec.rs`

I think it would be reasonable to make this an option (perhaps the default?), as it may interfere with vscode's ability to jump to source. However, this would drastically reduce line lengths when that functionality is not needed.

japaric:
this commit extends `--shorten-paths` so it now also handles:
- `/rustc` paths
- `rust-std` paths

it also adds highlighting to paths when `--shorten-paths` is not used

without `--shorten-paths`
Screenshot from 2021-05-20 14-34-44

with `--shorten-paths`
Screenshot from 2021-05-20 14-34-27

closes #66


src/dep/rust_std.rs
```diff
@@ -0,0 +1,117 @@
...
+            rust_repo_path: rust_repo::Path::One52(rust_repo::One52Path {
+                library: "library",
+                crate_name: "core",
+                path: StdPath::new("src/sync/atomic.rs"),
+            }),
+        };
+
+        assert_eq!(expected, path);
+
+        let expected_str = "[stable]/library/core/src/sync/atomic.rs";
```
japaric, line 112: the picture in the PR description doesn't show how a `rust-std` path is compressed but this test does

src/dep/rust_std/toolchain.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use rstest::rstest;
+
+    #[rstest]
+    #[case("stable-x86_64-unknown-linux-gnu", Channel::Stable, "stable")]
```
japaric, line 131: > the picture in the PR description doesn't show how a `rust-std` path is compressed but this test does

and this is how the toolchains are shortened with `--shorten-paths`

src/dep/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_std_path_returns_correct_variant() {
+        let cratesio = StdPath::new(
+        "/home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-rt-0.6.13/src/lib.rs",
+    );
```
Urhengulas, line 71: This seems a bit weirdly formatted 🤔

src/dep/rust_repo.rs
```diff
@@ -0,0 +1,122 @@
...
+            self.crate_name,
+            self.path.display(),
+            sep = path::MAIN_SEPARATOR
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::path::Path as StdPath;
```
Urhengulas, line 85: Isn't `StdPath` already imported with `use super::*`?

src/dep/cratesio.rs
```diff
@@ -0,0 +1,98 @@
+use std::path::{self, Component, Path as StdPath, PathBuf};
+
+use colored::Colorize as _;
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct Path<'p> {
```
Urhengulas, line 6: Why are you using `pub(crate)`? Could you not just use `pub`, since it is a application anyways?

Urhengulas (commented):
👍🏾

src/dep/mod.rs
japaric: indeed! that's what `rustfmt` is doing though so :shrug:

src/dep/rust_repo.rs
japaric: nicely spotted

src/dep/cratesio.rs
japaric: mainly out of habit: if it's not going to be used from outside the crate then I use `pub(crate)`. Also, rust-analyzer has a code assist to add `pub(crate)`, but not to add `pub` :wink:

japaric: > if it's not going to be used from outside the crate

in binary crates, the API is not meant to be used by other crates so I `pub(crate)` all the things

src/dep/mod.rs
Urhengulas: `rustfmt` seems to behave a bit weird here.. it allows me to also have the `let cratesio` in one line, but for `let rustc` it will format it into multiple 🤔 

see new suggestion 😉

src/dep/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        let rustc = StdPath::new(
+            "/rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/panicking.rs",
+        );
+        let rust_std = StdPath::new("/home/user/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/sync/atomic.rs");
```
Urhengulas, line 76:
```suggestion
        let cratesio = StdPath::new(
            "/home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-rt-0.6.13/src/lib.rs",
        );
        let rustc = StdPath::new(
            "/rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/panicking.rs",
        );
        let rust_std = StdPath::new(
            "/home/user/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/sync/atomic.rs",
        );
```

src/dep/cratesio.rs
Urhengulas: > in binary crates, the API is not meant to be used by other crates so I `pub(crate)` all the things

but using a application API is not possible even if it would be `pub`, is it? 👀

Urhengulas: > Also, rust-analyzer has a code assist to add `pub(crate)`, but not to add `pub` wink

I can see this reason 😉

Urhengulas:
Looks good to me and I like the shortening! 😄 

I think I would like to have this enabled per default and a flag for printing the long paths, but this is a story for another PR :D

japaric:
bors r=Urhengulas

PR #211, opened 2021-05-20
`mv backtrace.rs backtrace/mod.rs`

Urhengulas:


jonas-schievink:
Seems good to keep this consistent between knurling projects (defmt also uses `mod.rs`)

bors r+

PR #212, opened 2021-05-21
WIP: make `unwind::target()` infallible

Lotterleben:
in Draft status because my approach to error conversion feels clunky and I did resort to using a macro, so any feedback would be appreciated

### DIFF
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -1,6 +1,6 @@
 //! unwind target's program
 
-use anyhow::{ensure, Context as _};
+use anyhow::{anyhow, Context as _};
 use gimli::{
     BaseAddresses, DebugFrame, LittleEndian, UninitializedUnwindContext, UnwindSection as _,
 };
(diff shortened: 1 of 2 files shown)


jonas-schievink (commented):
I think this could be cleaned up a bit by moving the `loop` contents into its own function, which can then continue to use `?` to return early. The same goes for the register loading before the loop.

This might be easier if you put all required state in a struct, to avoid too many function arguments, but I haven't looked into this in detail.

jonas-schievink:
I thought you might be able to collect the `Outcome` outside of those functions, and have them early-return with an error

Lotterleben:
I don't think so– the outcome us updated subsequently as we're going through the loop. I'm checking if the update steps could be broken down into smaller functions that can then early return but they're all entangled in vakues that are created/updated earlier or later...

Lotterleben:
(That's probably another sign that this needs refactoring but makes a bit trickier than I initially thought :D)

jonas-schievink:
Yeah, I suppose it's fine to merge this as-is then

Lotterleben:
I've found one small helper function that could be extracted but other than that it gets messy

src/backtrace/unwind.rs
```diff
@@ -20,60 +20,67 @@ static MISSING_DEBUG_INFO: &str = "debug information is missing. Likely fixes:
 
 /// Virtually* unwinds the target's program
 /// \* destructors are not run
-// FIXME(?) this should be "infallible" and return as many frames as possible even in case of IO
-// errors
+///
+/// This returns as much info as could be collected, even if the collection is interrupted by an error.
+/// if an error occurred during processing, it is stored in `Output::processing_error`.
```
Urhengulas, line 25:
```suggestion
/// If an error occurred during processing, it is stored in `Output::processing_error`.
```

src/backtrace/unwind.rs
```diff
@@ -20,60 +20,67 @@ static MISSING_DEBUG_INFO: &str = "debug information is missing. Likely fixes:
...
+        ( $e:expr ) => {
+            match $e {
+                Ok(x) => x,
+                Err(err) => {
+                    output.processing_error = Some(anyhow!(err));
+                    return output;
+                }
+            }
+        };
+    }
```
Urhengulas, line 50: I generally don't like function, struct or macro definitions inside of functions and would vote for moving this below the function.

src/backtrace/unwind.rs
```diff
@@ -20,60 +20,67 @@ static MISSING_DEBUG_INFO: &str = "debug information is missing. Likely fixes:
...
-) -> anyhow::Result<Output> {
+) -> Output {
+    let mut output = Output {
+        corrupted: true,
+        outcome: Outcome::Ok,
+        raw_frames: vec![],
+        processing_error: None,
+    };
+
+    // returns all info collected until the error occurred and puts the error into `processing_error`
```
Urhengulas, line 39:
```suggestion
    /// Returns all info collected until the error occurred and puts the error into `processing_error`
```

src/backtrace/unwind.rs
```diff
@@ -123,28 +135,51 @@ pub(crate) fn target(
...
+    None
 }
 
 #[derive(Debug)]
 pub struct Output {
     pub(crate) corrupted: bool,
     pub(crate) outcome: Outcome,
     pub(crate) raw_frames: Vec<RawFrame>,
+    // will be `Some` if an error occured while putting together the output.
+    // `outcome` and `raw_frames` will contain all info collected until the error occurred.
```
Urhengulas, line 181:
```suggestion
    /// Will be `Some` if an error occured while putting together the output.
    /// `outcome` and `raw_frames` will contain all info collected until the error occurred.
```

src/backtrace/unwind.rs
```diff
@@ -20,60 +20,67 @@ static MISSING_DEBUG_INFO: &str = "debug information is missing. Likely fixes:
...
-            );
-
-            outcome = if overflowed_stack(sp, sp_ram_region) {
-                Outcome::StackOverflow
-            } else {
-                Outcome::HardFault
-            };
-        }
+        output.outcome = check_hard_fault(pc, vector_table, &mut output, sp, sp_ram_region)
+            .unwrap_or(output.outcome);
```
Urhengulas, line 64: The `.unwrap_or(output.outcome)` confused me at first, but it just mean that we just leave the value untouched if we return `None`?

Maybe if-let would be more explicit?

src/backtrace/unwind.rs
```diff
@@ -89,31 +96,36 @@ pub(crate) fn target(
...
-            raw_frames.push(RawFrame::Exception);
+            output.raw_frames.push(RawFrame::Exception);
 
             let fpu = match lr {
                 0xFFFFFFF1 | 0xFFFFFFF9 | 0xFFFFFFFD => false,
                 0xFFFFFFE1 | 0xFFFFFFE9 | 0xFFFFFFED => true,
-                _ => bail!("LR contains invalid EXC_RETURN value {:#010X}", lr),
+                _ => {
+                    output.processing_error =
+                        Some(anyhow!("LR contains invalid EXC_RETURN value {:#010X}", lr));
```
Urhengulas, line 113: (not necessarily for this PR): Since we use `output.processing_error = Some(anyhow!(...)` in multiple places we could have a small helper method for it 😄

Urhengulas (changes requested):

src/backtrace/unwind.rs
Lotterleben: I generally agree, the reason I kept this in scope of the function is because it is awful enough as it is and I wanted to prevent anyone else from using it

src/backtrace/unwind.rs
Lotterleben: duh! good point.done :D

src/backtrace/unwind.rs
Lotterleben: agreed, but I think this PR has overstayed its welcome :)

Lotterleben:
@Urhengulas are  you ok with the answers? I think you need to mark your approval somehow as the requests for changes are blocking currently

Urhengulas (approved):
> @Urhengulas are you ok with the answers? I think you need to mark your approval somehow as the requests for changes are blocking currently

Yes!

PR #215, opened 2021-05-27
WIP: probe-rs 0.11 update

Sheap:
In order to get probe-run working with the pre-production main branch of probe-rs, I had to make the following changes.

These changes appear to work, but I was just trying to quickly solve compiler issues, there may be better solutions, so I thought I'd open it up to the community.


jonas-schievink:
It just fails because probe-rs 0.11 has not yet been released to crates.io

Urhengulas:
> It just fails because probe-rs 0.11 has not yet been released to crates.io

Ups, didn't check and assumed it uses git dependency 🙈

Sheap:
Locally I am using local versions of probe-rs and probe-rs-rtt both checked out on master, and probe-rs-rtt further modified to also use my local probe-rs. It works (for some definition of works: see [here](https://github.com/nrf-rs/nrf-hal/issues/323), but I don't believe that issue is probe-run related)

Using git dependencies didn't work because of conflicts between probe-rs and probe-rs-rtt.

MathiasKoch:
Perhaps this could be revisited now that probe-rs 0.11.0 is released?
It would be nice to have the option of flash verification in probe-run.

Urhengulas:
Implemented in #235. Thank you @Sheap for caring about this!

PR #218, opened 2021-05-31
add first, user-triggered snapshot tests

Lotterleben:
to try them out locally, connect an nrf52840 and run `cargo test -- --ignored`

these test cases are pretty basic to give us some sort of baseline and can be expanded in subsequent PRs. In the future it'd be cool to have them [set up as hardware in the loop tests](https://ferrous-systems.com/blog/gha-hil-tests/) but I think they're already helpful during development as-is, even if we don't catch regressions automatically (yet)


tests/test.rs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            truncated_probe_run_output.push_str(line);
+            truncated_probe_run_output.push_str("\n");
+        }
+    }
+
+    truncated_probe_run_output
+}
+
+#[test]
```
Urhengulas, line 64: Could `rstest` be used to combine these tests into one with various cases? See https://docs.rs/rstest/0.10.0/rstest/#creating-parametrized-tests

Urhengulas (changes requested):
Love it! 😸

tests/test.rs
```diff
@@ -0,0 +1,105 @@
...
+            && !line.starts_with("   Compiling probe-run v")
+            // TODO don't drop these after https://github.com/knurling-rs/probe-run/issues/217 is resolved
+            && !line.starts_with("└─ ")
+        {
+            truncated_probe_run_output.push_str(line);
+            truncated_probe_run_output.push_str("\n");
+        }
+    }
+
+    truncated_probe_run_output
```
Urhengulas, line 61:
```suggestion
    probe_run_output
        .lines()
        .filter(|line| {
            !line.starts_with("    Finished")
            && !line.starts_with("     Running `")
            && !line.starts_with("    Blocking waiting for file lock ")
            && !line.starts_with("   Compiling probe-run v")
            // TODO don't drop these after https://github.com/knurling-rs/probe-run/issues/217 is resolved
            && !line.starts_with("└─ ")
        })
        .map(|line| format!("{}\n", line))
        .collect()
```

tests/test.rs
Urhengulas: (this obviously can be done in another PR)

tests/test.rs
```diff
@@ -0,0 +1,105 @@
...
+    handle.wait().unwrap();
+
+    // remove the lines printed during flashing, as they contain timing info that's not always the same
+    let mut truncated_probe_run_output = "".to_string();
+    for line in probe_run_output.lines() {
+        if !line.starts_with("    Finished")
+            && !line.starts_with("     Running `")
+            && !line.starts_with("    Blocking waiting for file lock ")
+            && !line.starts_with("   Compiling probe-run v")
+            // TODO don't drop these after https://github.com/knurling-rs/probe-run/issues/217 is resolved
```
Urhengulas, line 53: What are "these"? aka. let's maybe name it "paths" or something similar?

Urhengulas:
I am wondering if these tests will break with a next compiler release, because this might impact the program size, which is encoded e.g. here https://github.com/knurling-rs/probe-run/blob/7e4e05172959e7d0d307e62c4ada38f5ff05bf36/tests/snapshots/test__panic_is_reported_as_such.snap#L6

But it seems that it at least shows the same for the current nightly, so I guess this is a not-problem 😬

tests/test.rs
```diff
@@ -0,0 +1,105 @@
...
+use os_pipe::pipe;
+use std::io::Read;
+use std::sync::Mutex;
+use structopt::lazy_static::lazy_static;
+
+lazy_static! {
+    /// rust will try to run the tests in parallel by default, and `insta` doesn't like the
+    /// usual way of disabling this via `--test-threads=1`, so we're using this
+    /// mutex to make sure we're not re-flashing until the last run is finished
+    static ref ONE_RUN_AT_A_TIME: Mutex<i32> = Mutex::new(0i32);
```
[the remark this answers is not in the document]
Lotterleben: from what I gather, that's some [`lazy_static`](https://docs.rs/lazy_static/1.4.0/lazy_static/) trickery: We need the mutex to be "lazy static" because the Mutex is only created at runtime (whereas normally, only items that can be evaluated at compile time can be "regular" `static`). So our mutex is a [`ref`](https://doc.rust-lang.org/rust-by-example/scope/borrow/ref.html), e.g. a reference to a `Mutex` that is then  initialized when it is dereferenced for the first time (`lazy_static` implements this for us). Not 100% sure why this is required to be a `ref` rather than `static ONE_RUN_AT_A_TIME: &Mutex =  Mutex::new(0i32);` e.g. if that's either an implementation choice by `lazy_static` (and why it was made) or if that was the only way to go

tests/test.rs
Lotterleben: mhm, style-wise I personally don't think that would be a good idea for the *current* test cases, as they all check different functionalities and are pretty short. We'd be losing descriptiveness (test names as documentation) and wouldn't really win much in terms of deduplicating code. In other cases it might be  super useful though, e.g. if we're testing different values for `--max-backtrace-len`?

I'm not sure if/how that would work well with `insta`, as the snapshots carry the names of the test functions, but maybe that can be tweaked or insta is nice enough to encode any additional params in the name? Would be great to investigate in that future PR :)

tests/test.rs
Lotterleben: maybe it's me but I actually find that less readable=

tests/test.rs
Lotterleben: these are the `└─ probe_run @ /Users/lottesteenbrink/.cargo/registry/src/github.com-1ecc6299db9ec823/probe-run-0.2.3/src/main.rs:220` locations. I've made the copmment more clear

Lotterleben:
> I am wondering if these tests will break with a next compiler release, because this might impact the program size, which is encoded e.g. here
> 
> https://github.com/knurling-rs/probe-run/blob/7e4e05172959e7d0d307e62c4ada38f5ff05bf36/tests/snapshots/test__panic_is_reported_as_such.snap#L6
> 
> But it seems that it at least shows the same for the current nightly, so I guess this is a not-problem 😬

Since we're not re-compiling but running precompiled binaries that should be ok. If we ever recompile them then yes, we'll have to update the tests :D But I think that's actually a useful canary

tests/test.rs
Urhengulas: I personally prefer to use iterators, cause they are more compact and you need no mutable working variables. For the readability: By now I use it a lot and think it is very clear what it is doing, but maybe that is not true if you are not used to them.

🤷🏾‍♀️

tests/test.rs
Urhengulas: 👍🏾

tests/test.rs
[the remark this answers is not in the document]
Urhengulas: Thank you for the elaboration! 😄

tests/test.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// run probe-run with `args` and truncate the "Finished .. in .." and "Running `...`" flashing output
+/// NOTE: this currently only capures `stdin`, so any `log::` ed output, like flashing
+fn run_and_truncate(args: &str) -> String {
+    let _guard = ONE_RUN_AT_A_TIME.lock().unwrap();
+
+    // add prefix to run this repository's version of `probe-run` and
+    // remove user-dependent registry and rustc information from backtrace paths
+    let complete_command = format!("run -- {} --shorten-paths", args);
+    let complete_command_args = complete_command.split(" ");
```
Urhengulas, line 21: If you don't like to integrate it into the command as suggested, you could also combine these two lines:
```suggestion
    let complete_command_args = format!("run -- {} --shorten-paths", args).split(" ");
```

tests/test.rs
```diff
@@ -0,0 +1,105 @@
+use insta;
+use os_pipe::pipe;
+use std::io::Read;
+use std::sync::Mutex;
+use structopt::lazy_static::lazy_static;
```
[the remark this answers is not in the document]
Urhengulas: Don't you like it?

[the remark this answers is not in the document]
Urhengulas: At least the `use insta` is unnecessary.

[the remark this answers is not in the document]
Lotterleben: ah, you're right

tests/test.rs
```diff
@@ -0,0 +1,100 @@
...
+        .stdout(writer)
+        .stderr(writer_clone)
+        // run `probe-run`
+        .spawn()
+        .unwrap();
+
+    // retrieve output and clean up
+    let mut probe_run_output = String::new();
+    reader.read_to_string(&mut probe_run_output).unwrap();
+    handle.wait().unwrap();
```
japaric, line 38: maybe be worthwhile to check and return the exit code of `probe-run` from here and assert that it's zero or non-zero in each unit test (e.g. panic should be non-zero exit, etc.)

tests/test.rs
Lotterleben: fair point about the non-mutable interim var

tests/test.rs
```diff
@@ -0,0 +1,100 @@
...
+        // capture stderr and stdout while preserving line order
+        .stdout(writer)
+        .stderr(writer_clone)
+        // run `probe-run`
+        .spawn()
+        .unwrap();
+
+    // retrieve output and clean up
+    let mut probe_run_output = String::new();
+    reader.read_to_string(&mut probe_run_output).unwrap();
```
japaric, line 37: the order of these operations seemed a bit unusual: I would wait for the process to finish and then read the pipe but here it's done the other way -- I guess `read_to_string` blocks until the two `writer` handles are closed so you don't see "incomplete data"?

out of curiosity can you do `status` (to drive the program to completion) instead of `spawn` and then read out from the `reader`? or does that not work due to how pipes work (it may be an error to read from the `reader` end when the `writer` end has been closed?)

japaric (approved):
other than the inline comments, this looks good!

tests/test.rs
Lotterleben: if we use `status()` (without `wait()` in because there's no child to wait for) the tests turn flaky; the command hangs in some cases but not all and it's not reproducible which and how many will pass. The same happens if we move `handle.wait()` before the string reading

Urhengulas (approved):

japaric:
@Lotterleben feel free to send this to bors after a rebase

PR #219, opened 2021-05-31
add more explicit hint if elf path doesn't lead to an existing file

Lotterleben:
small ux improvement

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -143,6 +143,12 @@ fn notmain() -> anyhow::Result<i32> {
     let max_backtrace_len = opts.max_backtrace_len;
     let shorten_paths = opts.shorten_paths;
     let elf_path = opts.elf.as_deref().unwrap();
+    if !elf_path.exists() {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -143,6 +143,12 @@ fn notmain() -> anyhow::Result<i32> {
     let max_backtrace_len = opts.max_backtrace_len;
     let shorten_paths = opts.shorten_paths;
     let elf_path = opts.elf.as_deref().unwrap();
+    if !elf_path.exists() {
+        return Err(anyhow!(
+            "Can't find ELF file at `{}`; are you sure you got the right path?",
```
jonas-schievink, line 148: Error messages are usually lowercased:

```suggestion
            "can't find ELF file at `{}`; are you sure you got the right path?",
```

Other than that, seems good, feel free to r+

Lotterleben:
welp, chaos lotter strikes again

PR #221, opened 2021-06-01
Obtain git-version from macro, instead of custom build-script

Urhengulas:
This implements the idea from @japaric to utilize the `git_version` crate for obtaining the git hash, instead of the custom build-script we used so far.

In case `git` is not present on the system the reported version will only consist of the `CARGO_PKG_VERSION`. 

---

_note: I am a bit unsure where exactly to put the `extract_git_hash` helper function. I mainly considered the `utils` module. I decided to put it into `main.rs` for now to not potentially conflict with the upcoming refactoring from @Lotterleben and @japaric._


Lotterleben:
main.rs will be heavily refactored, *but* this PR shouldn't be too painful to rebase onto the refactored  code I think. If that*#s okay with you, I'd like to wait with merging this PR until after the refactoring is done.

Urhengulas:
> main.rs will be heavily refactored, *but* this PR shouldn't be too painful to rebase onto the refactored  code I think. If that*#s okay with you, I'd like to wait with merging this PR until after the refactoring is done.

Sure!

Urhengulas:
@Lotterleben I've rebased it onto the latest main.

src/cli.rs
```diff
@@ -122,12 +122,51 @@ fn print_chips() {
...
+    git_describe.split("-").nth(2).unwrap()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn should_extract_hash_from_description() {
+        let hash = extract_git_hash("v0.2.3-12-g25c50d2");
```
Lotterleben, line 157: what happens if there are no extra commits yet after a fresh tag, e.g. a `git describe` yields output like `v0.2.3`?

Urhengulas: It won't only give us the tag, but the long format, since we specify `args = ["--long"]`.

```
       --long
           Always output the long format (the tag, the number of commits and
           the abbreviated commit name) even when it matches a tag. This is
           useful when you want to see parts of the commit object name in
           "describe" output, even when the commit in question happens to be a
           tagged version. Instead of just emitting the tag name, it will
           describe such a commit as v1.2-0-gdeadbee (0th commit since tag
           v1.2 that points at object deadbee....).
```

Lotterleben: aaah perfect

Lotterleben (approved):

Lotterleben:
feel free to bors rplus

PR #222, opened 2021-06-04
refactor the huge "main" function into smaller functions + module

japaric:
what the PR title says. I ran the ELFs from PR #218 and they print the expected output.

This is the most non-trivial change that may have changed semantics
- before we were extracting *two* RAM regions from the probe-rs registry. The "first one" (returned by the iterator) and the "one that contains the initial stack pointer (initial_sp)". we installed the canary on "the first one" but only if it contained the initial_sp
- now, we only extract one RAM region, the one that contains initial_sp, and install the canary there

I think this still does the same, in practice, but not 100% sure.

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -2,7 +2,7 @@ use std::path::Path;
 
 use probe_rs::{config::RamRegion, Core};
 
-use crate::sketch::ProcessedElf;
+use crate::elf::ProcessedElf;
 
 mod pp;
 mod symbolicate;
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -2,7 +2,7 @@ use std::path::Path;
 
 use probe_rs::{config::RamRegion, Core};
 
-use crate::elf::ProcessedElf;
+use crate::elf::Elf;
 
 mod pp;
 mod symbolicate;
@@ -18,7 +18,7 @@ pub(crate) struct Settings<'p> {
 /// (virtually) unwinds the target's program and prints its backtrace
 pub(crate) fn print(
     core: &mut Core,
-    elf: &ProcessedElf,
+    elf: &Elf,
     active_ram_region: &Option<RamRegion>,
     settings: &Settings,
 ) -> anyhow::Result<Outcome> {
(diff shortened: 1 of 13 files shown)


src/backtrace/mod.rs
```diff
@@ -62,3 +53,37 @@ pub(crate) fn print(
...
+                log::error!("the program panicked");
+            }
+            Outcome::Ok => {
+                log::info!("device halted without error");
+            }
+        }
+    }
+}
+
+/// convert outomce into an exit code
```
jonas-schievink, line 81:
```suggestion
/// Converts `Outcome` to an exit code.
```

src/canary.rs
```diff
@@ -0,0 +1,108 @@
...
+/// |        |
+/// |        |
+/// +--------+
+/// | canary |
+/// +--------+ -> highest_static_var_address
+/// |        |
+/// | static | (variables, fixed size)
+/// |        |
+/// +--------+ -> lowest RAM address
+/// ```
```
jonas-schievink, line 28: Awesome :heart:

src/canary.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// The whole canary is initialized to `CANARY_VALUE` before the target program is started.
+/// The canary size is 10% of the available stack space or 1 KiB, whichever is smallest.
+///
+/// When the programs ends (due to panic or breakpoint) the integrity canary is checked. If it was
```
jonas-schievink, line 33:
```suggestion
/// When the programs ends (due to panic or breakpoint) the integrity of the canary is checked. If it was
```

src/canary.rs
```diff
@@ -0,0 +1,108 @@
...
+/// "touched" (any of its bytes != `CANARY_VALUE`) then that is considered to be a *potential* stack
+/// overflow
```
jonas-schievink, line 35:
```suggestion
/// overflow.
```

src/canary.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// The canary is not installed if the program memory layout is "inverted" (stack is *below* the
+/// static variables)
```
jonas-schievink, line 38:
```suggestion
/// static variables).
```

jonas-schievink (commented):
This looks like a very nice improvement, nice!

I haven't thoroughly reviewed every part of this, but if the test-ELFs work then I'm fine with merging this.

japaric:
bors r=jonas-schievink

PR #224, opened 2021-06-07
target_info: print ram region again

Lotterleben:
after the [snapshot test](https://github.com/knurling-rs/probe-run/pull/218/files) rebase I noticed we're not printing the RAM region anymore. I'm not sure if I'm misremembering that that was on purpose. If I *am* misremembering, this PR fixes that. If I'm not misremembering, it leaves a paper trail.


japaric (approved):
looks good to me. if's debug statement so not much harm in having it since it's not printed by default

PR #225, opened 2021-06-14
`cli::tests`: rstest-ify tests for `fn extract_git_hash`

Urhengulas:


jonas-schievink (commented):
What is the benefit of this over just calling a helper like this 3 times?

```rust
fn check(description: &str, expected: &str) {
    assert_eq!(extract_git_hash(description), expected);
}
```

This is so short that even the helper method might be unnecessary

Urhengulas:
> What is the benefit of this over just calling a helper like this 3 times?

These are the 3 potential input values which are returned by `git_version!(fallback = "--", args = ["--long"])`.

> ```rust
> fn check(description: &str, expected: &str) {
>     assert_eq!(extract_git_hash(description), expected);
> }
> ```

If you want you I can combine line 162 and 163 into one line.

> This is so short that even the helper method might be unnecessary

What do you mean? That it would be unnecessary to use `rstest`? I prefer it, since it will be three tests and I will know which one failed, instead of knowing that one of them failed.

jonas-schievink:
You could also write 3 individual test functions, they'd all be one-liners. I am somewhat skeptical that a proc macro can make this any better.

Urhengulas:
> You could also write 3 individual test functions, they'd all be one-liners. I am somewhat skeptical that a proc macro can make this any better.

I mean that is how it is currently (on main). It's just a bit more compact with `rstest`.

Urhengulas:
ping 🔔

jonas-schievink:
I still really don't think that rstest pulls its weight here, but since we're already using it and I don't really want to spend more time on this, bors r+

Urhengulas:
> I still really don't think that rstest pulls its weight here, but since we're already using it and I don't really want to spend more time on this, bors r+

I'd be interested why you are so sceptical? From my user perspective I am able to combine three test functions into one, but still get the same test behavior.

jonas-schievink:
I think replacing 3 very simple functions with a niche proc macro significantly increases the effort needed to understand them (since you now need to understand what the macro does). If you have a large code base with lots of tests that get simplified by it, then it might be worth it, but I don't think probe-run is such a code base.

Urhengulas:
> I think replacing 3 very simple functions with a niche proc macro significantly increases the effort needed to understand them (since you now need to understand what the macro does). If you have a large code base with lots of tests that get simplified by it, then it might be worth it, but I don't think probe-run is such a code base.

Thank you for the elaboration 🙂

Issue #239, opened 2021-07-09
add troubleshooting info to "probe not found" error


japaric:
`probe-run` could print the info I posted in https://github.com/knurling-rs/probe-run/issues/238#issuecomment-877260800 so users can troubleshoot / debug the issue on their own

Urhengulas:
We can additionally add it to the Troubleshooting section of the README. And maybe put it to a separate file, since it got quite long by now.

japaric:
>  the Troubleshooting section of the README

I forgot we had that! I think we can link to that from the error message. It should use a more stable link (commit hash or pinned to a version) than e.g. https://github.com/knurling-rs/probe-run#troubleshooting

Lotterleben:
I understand why we'd want a more stable link (e.g. if someone is using an old probe-run version, newer info might not help), but how do we ensure we remember to bump this link regularly?

Urhengulas:
> It should use a more stable link (commit hash or pinned to a version) than e.g. https://github.com/knurling-rs/probe-run#troubleshooting

I think linking to the headline of the troubleshooting section should be fine, since this won't change too often. If it would be a link to a sub-(sub-)headline I'd agree.

PR #247, opened 2021-08-16
Print troubleshooting information on "probe not found" error

Lotterleben:
resolves https://github.com/knurling-rs/probe-run/issues/239 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,17 +201,38 @@ $ cargo run --bin hello --force-backtrace
 
 ## Troubleshooting
 
-### `probe-run --list-probes` says "No devices were found."
+### "Error: no probe was found."
 
-Apart from a faulty connection between your computer and the target device, this could be caused by several things:
+First, check your data cable:
+
+- make sure that it is connected to the right port on your development board
+- make sure that you are using a **data** cable– some cables are built for charging only! When in doubt, try using a different cable.
+
+If this doesn't resolve the issue, try the following:
 
 #### [Linux only] udev rules haven't been set
 
-In order for `probe-run` to find the device you'd like to run your code on, your system needs permission to access the device as a non-root user.
+Check if your device shows up in `lsusb`:
+
+```console
+$ lsusb
+Bus 001 Device 008: ID 1366:1015 SEGGER J-Link
+# if your device shows up like this ^^^^^^^^, skip to the next troubleshooting section
+```
+
+**If it doesn't show up**, you need to give your system permission to access the device as a non-root user so that `probe-run` can find your device.
 
 In order to grant these permissions, you'll need to add a new set of udev rules.
 
 To learn how to do this for the nRF52840 Development Kit, check out the [installation instructions](https://embedded-trainings.ferrous-systems.com/installation.html?highlight=udev#linux-only-usb) in our embedded training materials.
+
+afterwards, your device should show up in `probe-run --list-probes` similar to this:
+
+```console
+$ probe-run --list-probes
+The following devices were found:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -202,18 +201,39 @@ $ cargo run --bin hello --force-backtrace
...
 #### [Linux only] udev rules haven't been set
 
-In order for `probe-run` to find the device you'd like to run your code on, your system needs permission to access the device as a non-root user.
+Check if your device shows up in `lsusb`:
+
+```console
+$ lsusb
+Bus 001 Device 008: ID 1366:1015 SEGGER J-Link
+# if your device shows up like this ^^^^^^^^, skip to the next troubleshooting section
+```
```
Urhengulas, line 221:
````suggestion
```

If your device shows up like in the example, skip to the next troubleshooting section
````

File: src/probe.rs [hunk 2]
```diff
@@ -5,6 +5,10 @@ use probe_rs::{DebugProbeInfo, Probe};
 
 use crate::cli;
 
+const NO_PROBE_FOUND_ERR: &str = "no probe was found.\n
+Common reasons for this are faulty cables or missing permissions.
+For detailed instructions, visit: https://github.com/knurling-rs/probe-run/tree/2f138c3#troubleshooting";
```
Urhengulas, line 10: Can't we link to the main branch troubleshooting section?

```suggestion
For detailed instructions, visit: https://github.com/knurling-rs/probe-run#troubleshooting";
```

src/probe.rs
```diff
@@ -15,7 +19,7 @@ pub(crate) fn open(opts: &cli::Opts) -> Result<Probe, anyhow::Error> {
     };
 
     if filtered_probes.is_empty() {
-        bail!("no probe was found")
+        bail!("{}", NO_PROBE_FOUND_ERR)
```
Urhengulas, line 22: Why is there no leading "Error" as in line 50?

Urhengulas (changes requested):

README.md
```diff
@@ -202,18 +201,39 @@ $ cargo run --bin hello --force-backtrace
...
 
 #### [Linux only] udev rules haven't been set
 
-In order for `probe-run` to find the device you'd like to run your code on, your system needs permission to access the device as a non-root user.
+Check if your device shows up in `lsusb`:
+
+```console
+$ lsusb
+Bus 001 Device 008: ID 1366:1015 SEGGER J-Link
+# if your device shows up like this ^^^^^^^^, skip to the next troubleshooting section
```
Urhengulas, line 220: What are common debuggers, except "J-link"? Maybe we can give them a few keywords to scan for, in case they have multiple USB devices and don't know if one is their debugger or not.

src/probe.rs
Urhengulas: Just saw the discussion in the issue

src/probe.rs
Lotterleben: Because `bail!()` adds the "Error: " prefix automatically (that's why I added it in line 50 as well, for consistency)

README.md
Lotterleben: afaik J-Links are by far the most common. I'd rather not add lots of text by trying to create a comprehensive list (in my experience, the longer the tutorial the more likely it won't be thoroughly read and the info you're putting in there is lost in most cases)
If you're unsure, you can always unplug, `lsusb`, plug in, `lsusb`, and look at the difference. (I could add that as a tip?)

Urhengulas: Yes, I think this is a useful tip!

Issue #198, opened 2021-05-12
change CLI around --force-backtrace and --max-backtrace-len


japaric:
We are going to change the CLI around `--force-backtrace` and `--max-backtrace-len`. The accepted proposal is https://github.com/knurling-rs/probe-run/issues/198#issuecomment-889073456 . This is only waiting on a PR to implement that proposal.

---

Original ticket:

(this is technically a breaking change so it should wait until v0.3.0)

`--max-backtrace-len=0` is currently allowed. While working on PR #197 it wasn't clear to me what the semantics of `--force-backtrace --max-backtrace-len=0` should be but I made `--max-backtrace-len=0` never print a backtrace regardless of other flags and the presence of stack overflows or exceptions.

IMO, it'd be better to reject `--max-backtrace-len=0` and only accepts lengths equal or greater than 1; that avoids the above ambiguity.

cc @Lotterleben

Lotterleben:
makes sense. I'm wondering, is there a case where one might not want any backtraces ever?

japaric:
> I'm wondering, is there a case where one might not want any backtraces ever?

we also print the backtrace to stdout, right? perhaps if you want to grep for something in the logs but *not* in the backtrace then it may be useful to completely disable the backtrace. this use case would be best served by the proposed JSON output (#178)

jonas-schievink:
IMO it makes sense to reject `--force-backtrace --max-backtrace-len=0`, but I don't think `--max-backtrace-len=0` by itself should be rejected. I would expect it to just turn off the backtrace (either completely, with no output referring to the backtrace, or in a way that just prints a backtrace with 0 frames).

japaric:
> but I don't think --max-backtrace-len=0 by itself should be rejected. I would expect it to just turn off the backtrace

that sounds like 'some way to disable the backtrace' is a desirable thing.

perhaps we can harmonize this to work like Cargo's `--color` flag? `--backtrace={auto,always,never}`. `auto` is the current, default behavior (e.g. backtrace on panic, but not on `exit`). `always` is `--force-backtrace` (which is no longer needed with this scheme). `never` disables the backtrace completely.

THEN `--max-backtrace-len` can be limited to be greater than 1 and shall only be used with `--backtrace={auto,always}`. e.g. `--max-backtrace-len=10 --backtrace=never` would be an error.

[...]

PR #250, opened 2021-08-18
Backtrace options

BriocheBerlin:
Closes https://github.com/knurling-rs/probe-run/issues/198


tests/test.rs
```diff
@@ -75,17 +75,6 @@ fn successful_run_has_no_backtrace() {
     insta::assert_snapshot!(run_result.output);
 }
 
-#[test]
-// this test should not be run by default, as it requires the target hardware to be present
-#[ignore]
-fn successful_run_can_enforce_backtrace() {
```
Lotterleben, line 81: why is this deleted and not adapted?

src/backtrace/mod.rs
```diff
@@ -57,10 +48,18 @@ pub(crate) fn print(
...
+    let print_backtrace= match settings.backtrace {
+        BacktraceOptions::Never => false,
+        BacktraceOptions::Always => true,
+        BacktraceOptions::Auto => if settings.panic_present
+            || unwind.outcome == Outcome::StackOverflow
+            || unwind.corrupted
+            || contains_exception {
+                true
+            } else {
+                false
```
Lotterleben, line 60: this `if...else` can be left out because the `settings.panic_present || unwind.outcome == Outcome::StackOverflow || ...` statement itself evaluates to a `bool`, which is what we want to return. e.g.:

```rust
 BacktraceOptions::Auto => settings.panic_present
             || unwind.outcome == Outcome::StackOverflow
             || unwind.corrupted
             || contains_exception 
```
should work as well

src/backtrace/mod.rs
```diff
@@ -57,10 +48,18 @@ pub(crate) fn print(
         .iter()
         .any(|raw_frame| raw_frame.is_exception());
 
-    let print_backtrace = settings.backtrace
-        || unwind.outcome == Outcome::StackOverflow
-        || unwind.corrupted
-        || contains_exception;
+    let print_backtrace= match settings.backtrace {
```
Lotterleben, line 51: ha, neat!

tests/test.rs
BriocheBerlin: Didn't realize at first that we keep the functionality (under a different name). Adapted!

src/backtrace/mod.rs
```diff
@@ -8,10 +8,29 @@ mod pp;
...
 mod unwind;
 
+#[derive(PartialEq, Eq)]
+pub(crate) enum BacktraceOptions {
+    Auto,
+    Never,
+    Always,
+}
+
+impl From<&String> for BacktraceOptions {
```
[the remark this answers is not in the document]
BriocheBerlin: Thanks, will leave (merge) for now but probably implement later.

PR #253, opened 2021-08-29
feat: add help message for JtagNoDeviceConnected

numero-744:
Hello!

I had the `JtagNoDeviceConnected` error and did not know what to do, and someone in the Rust Embedded matrix channel told me that it happens to a lot of people. So maybe `probe-run` could print some help in this case. This is what this PR does.

```
% cargo run --bin=blinky
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `/home/*****/Development/probe-run/target/debug/probe-run --chip STM32F407VGTx target/thumbv7em-none-eabi/debug/blinky`
Info: Jtag cannot find a connected device.
Help:
    Try using probe-run with option `--connect-under-reset`
    Or if using cargo:
        cargo run -- --connect-under-reset
    This error comes from the program currently in the chip and
    using `--connect-under-reset` is only a workaround.

Error: An error specific to a probe type occured

Caused by:
    Command failed with status JtagNoDeviceConnected
```

This PR is just to show the idea, I can refactor it as needed.

In the matrix channel was also discussed the possibility to:

- enable the flag automatically when the error occurs, which can be done while printing a warning, and/or
- add a link to an explanation of the reason why this error occurs and how to avoid it in the future

Refactoring of this PR would enable to add help for the user about more various kinds of errors.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -63,11 +63,13 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
             if e.to_string().contains("JtagNoDeviceConnected") {
                 eprintln!("Info: Jtag cannot find a connected device.");
                 eprintln!("Help:");
-                eprintln!("    Try using probe-run with option `--connect-under-reset`");
-                eprintln!("    Or if using cargo:");
+                eprintln!("    Check that the debugger is connected to the chip, if so");
+                eprintln!("    try using probe-run with option `--connect-under-reset`");
+                eprintln!("    or, if using cargo:");
                 eprintln!("        cargo run -- --connect-under-reset");
-                eprintln!("    This error comes from the program currently in the chip and");
-                eprintln!("    using `--connect-under-reset` is only a workaround.\n");
+                eprintln!("    If using this flag fixed your issue, this error might");
+                eprintln!("    come from the program currently in the chip and using");
+                eprintln!("    `--connect-under-reset` is only a workaround.\n");
             }
         }
         probe_attach?
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,6 +60,8 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     } else {
         let probe_attach = probe.attach(probe_target);
         if let Err(probe_rs::Error::Probe(ProbeSpecific(e))) = &probe_attach {
+            // FIXME Using `to_string().contains(...)` is a workaround as the concrete type
+            // of `e` is not public and therefore does not allow downcasting.
             if e.to_string().contains("JtagNoDeviceConnected") {
                 eprintln!("Info: Jtag cannot find a connected device.");
                 eprintln!("Help:");
--- a/src/main.rs
+++ b/src/main.rs
@@ -59,8 +59,8 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
         probe.attach_under_reset(probe_target)?
     } else {
         let probe_attach = probe.attach(probe_target);
-        if let Err(probe_rs::Error::Probe(ProbeSpecific(ref e))) = probe_attach {
-            if format!("{}", e).contains("JtagNoDeviceConnected") {
+        if let Err(probe_rs::Error::Probe(ProbeSpecific(e))) = &probe_attach {
+            if e.to_string().contains("JtagNoDeviceConnected") {
                 eprintln!("Info: Jtag cannot find a connected device.");
                 eprintln!("Help:");
                 eprintln!("    Try using probe-run with option `--connect-under-reset`");
--- a/src/main.rs
+++ b/src/main.rs
@@ -24,6 +24,7 @@ use colored::Colorize as _;
 use defmt_decoder::Locations;
 use probe_rs::{
     flashing::{self, Format},
+    DebugProbeError::ProbeSpecific,
     MemoryInterface as _, Session,
 };
 use probe_rs_rtt::{Rtt, ScanRegion, UpChannel};
@@ -57,7 +58,19 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let mut sess = if opts.connect_under_reset {
         probe.attach_under_reset(probe_target)?
     } else {
-        probe.attach(probe_target)?
+        let probe_attach = probe.attach(probe_target);
+        if let Err(probe_rs::Error::Probe(ProbeSpecific(ref e))) = probe_attach {
+            if format!("{}", e).contains("JtagNoDeviceConnected") {
+                eprintln!("Info: Jtag cannot find a connected device.");
+                eprintln!("Help:");
+                eprintln!("    Try using probe-run with option `--connect-under-reset`");
+                eprintln!("    Or if using cargo:");
+                eprintln!("        cargo run -- --connect-under-reset");
+                eprintln!("    This error comes from the program currently in the chip and");
+                eprintln!("    using `--connect-under-reset` is only a workaround.\n");
+            }
+        }
+        probe_attach?
     };
     log::debug!("started session");
 


src/main.rs
```diff
@@ -57,7 +58,19 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let mut sess = if opts.connect_under_reset {
         probe.attach_under_reset(probe_target)?
     } else {
-        probe.attach(probe_target)?
+        let probe_attach = probe.attach(probe_target);
+        if let Err(probe_rs::Error::Probe(ProbeSpecific(ref e))) = probe_attach {
+            if format!("{}", e).contains("JtagNoDeviceConnected") {
```
Urhengulas, line 63:
```suggestion
        if let Err(probe_rs::Error::Probe(ProbeSpecific(e))) = &probe_attach {
            if e.to_string().contains("JtagNoDeviceConnected") {
```

Urhengulas (changes requested):
Hi @numero-744, 

Thank you for your patch!
Is it possible to reproduce this error with the nRF52 DK? If so, could you please tell me how so I can test it? 😁 

> add a link to an explanation of the reason why this error occurs and how to avoid it in the future

This explaination could very well go into our troubleshooting section (https://github.com/knurling-rs/probe-run#troubleshooting). Do you want to add it as part of this PR?

> enable the flag automatically when the error occurs, which can be done while printing a warning

This also sounds good, but probably rather done in another PR. What do you think? Do you see a situation where enabling automatically could be disadvantageous aka. breaking things?

numero-744:
Thanks for your reply!

My only board is an stm32 so I couldn't tell… Maybe try to flash a program which loops a `WFI`, but not sure it disables the debug like my stm32 does.

For the explanation I agree for the place, I could be able to add a link in the printed text but I do not have the knowledge to explain precisely what is the problem and why it is… All I say here is what I have understood of the matrix discussion so well, no guarantee or expertise from me about this topic :grin:

For the automatic use of the flag I do not know if it would break things, I wanted to discuss about it in this thread. But it could be in another PR as well. It is up to you, just tell me your preference.

Anyway, I will commit your suggestions, and I'd like to refactor a bit this PR to avoid putting all this text into `main.rs`. No precise idea for now, suggestions are welcome! I think it would be great to have a place (module?) to put everything that could help users facing errors: diagnose the error, print warnings and suggestions, and if possible take the needed action. For instance when I used `probe-run` for the first time the ST-Link needed an update and I did not know what "firmware" `probe-run` was telling me about. The explanation and suggestion for this case could be added in the same module.

PS: thanks for `e.to_string()`, I do not know why I had not found this method in the docs earlier!

src/main.rs
```diff
@@ -57,7 +58,19 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
...
-        probe.attach(probe_target)?
+        let probe_attach = probe.attach(probe_target);
+        if let Err(probe_rs::Error::Probe(ProbeSpecific(e))) = &probe_attach {
+            if e.to_string().contains("JtagNoDeviceConnected") {
+                eprintln!("Info: Jtag cannot find a connected device.");
+                eprintln!("Help:");
+                eprintln!("    Try using probe-run with option `--connect-under-reset`");
+                eprintln!("    Or if using cargo:");
+                eprintln!("        cargo run -- --connect-under-reset");
+                eprintln!("    This error comes from the program currently in the chip and");
```
jonas-schievink, line 69: We shouldn't state this with certainty, it is possible that there really just isn't a device connected

src/main.rs
```diff
@@ -57,7 +58,19 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let mut sess = if opts.connect_under_reset {
         probe.attach_under_reset(probe_target)?
     } else {
-        probe.attach(probe_target)?
+        let probe_attach = probe.attach(probe_target);
+        if let Err(probe_rs::Error::Probe(ProbeSpecific(e))) = &probe_attach {
+            if e.to_string().contains("JtagNoDeviceConnected") {
```
jonas-schievink, line 63: It would be better to downcast to the concrete error type and inspect that. If probe-rs doesn't expose that type, please add a `// FIXME` comment explaining that.

src/main.rs
numero-744: Maybe we should add something like "Check that the debugger is connected to the chip, if so try […]" and then "If this flag fixed your issue, this error might come from […]"?

src/main.rs
numero-744: `e` is `Box<dyn std::error::Error + Send + Sync>`, can I downcast it?

`dbg!` tells me

```
e = CommandFailed(
    JtagNoDeviceConnected,
)
```

And using `cargo doc` I cannot find `CommandFailed` nor `JtagNoDeviceConnected`.

Urhengulas: Searching through the repo I found the `enum StlinkError::CommandFailed(...)` ([link](https://github.com/probe-rs/probe-rs/blob/b7e7cfd667a0a85fe012017f3cf865038001970e/probe-rs/src/probe/stlink/mod.rs#L1122)). This is only `pub(crate)` and therefore not accessible for us. If I understand it correctly we therefore can't downcast it.

@jonas-schievink What do you think is the best way to move forward? Ask them to expose the type?

src/main.rs
Urhengulas: > "Check that the debugger is connected to the chip, if so try […]" and then "If this flag fixed your issue, this error might come from […]"?

That sounds good

src/main.rs
jonas-schievink: > Ask them to expose the type?

Yes, that is needed to implement this more cleanly.

For now, we can use the `.to_string()` solution, since this is just a diagnostics improvement anyways. There should be at least a `// FIXME` comment in the code explaining this though.

numero-744: Comment suggestion:

```rust
// FIXME Using `to_string().contains(...)` is a workaround as the concrete type of
// `e` is not public and therefore does not allow downcasting.
```

numero-744:
About refactoring, I was thinking about adding `src/troubleshooting.rs` defining `pub trait Diagnoseable { fn diagnose(self) -> Self) }` and implementing it for `Result<probe_rs::Session, probe_rs::Error>` using a `match &self` and returning `self`. Actually it works but I think it would be suitable only to print info messages, not to recover from errors. It might be a problem as, if error recovering (for instance in this case use `.attach_under_reset(...)` instead of `.attach(...)`) is added later, the logic of error identification will be duplicated. What do you think about it?

Below is an example implementation of the suggestion above.

```rust
// src/main.rs:61
probe.attach(probe_target).diagnose()?
```

```rust
// src/troubleshooting.rs
use probe_rs::{DebugProbeError, Error, Session};

/// This trait is to be applied to results that can lead to warning messages helping the user to
/// fix the issue that raised an error
pub trait Diagnoseable {
    fn diagnose(self) -> Self;
}

impl Diagnoseable for Result<Session, Error> {
    fn diagnose(self) -> Self {
        match &self {
            // Ok should never be diagnosed
            Ok(_) => (),
            // Add some Err(...) branches here to manage the different kinds of errors
            Err(Error::Probe(DebugProbeError::ProbeSpecific(e))) => {
                // FIXME Using `to_string().contains(...)` is a workaround as the concrete type of
                // `e` is not public and therefore does not allow downcasting.
                if e.to_string().contains("JtagNoDeviceConnected") {
                    print_jtag_no_device_connected_info()
                }
            }
            // The other kinds of error are not managed yet, or there is nothing to be done for
            // them
            Err(_) => (),
        }
        self
    }
}

fn print_jtag_no_device_connected_info() { ... }
```

Urhengulas:
I've opened https://github.com/probe-rs/probe-rs/issues/797 to address the FIXME.

Urhengulas:
> About refactoring, I was thinking about adding `src/troubleshooting.rs` defining `pub trait Diagnoseable { fn diagnose(self) -> Self) }` and implementing it for `Result<probe_rs::Session, probe_rs::Error>` using a `match &self` and returning `self`. Actually it builds but I think it would be suitable only to print info messages, not to recover from errors. It might be a problem as, if error recovering (for instance in this case use `.attach_under_reset(...)` instead of `.attach(...)`) is added later, the logic of error identification will be duplicated. What do you think about it?

I guess think this is a good optimization, in case we have multiple variants of that error. For now I don't think it is necessary.

Let's wrap the PR as it is. Feel free to open an issue with your proposed `trait Diagnoseable`, so we can discuss it there.

bors r+

Issue #188, opened 2021-04-29
flag to report observed max stack usage


japaric:
- what: add an opt-in flag to `probe-run` that: fills the stack with a known bit pattern and then reports the "watermark level" of stack usage after the program exits

- how:
  - after reset halt and before flashing, fill the whole RAM region *that contains the SP* with a known 32-bit pattern
  - once the program exits and there wasn't a stack overflow, scan the RAM from the lowest address to its highest address for the first non-occurrence of the 32-bit pattern. tip: do a binary search
  - subtract this value from 'initial SP value' and report this as the *observed* max stack usage in bytes

- good to know:
  - (iirc) the ARM ABI says that the stack pointer must always be 8 byte aligned on function entry and from looking at machine code the stack pointer is always kept at a multiple of 4, hence the 32-bit pattern above
  - to paint the stack you want to do a `memset` operation at the SWD level to avoid moving, e.g. 256KB of repeated data, from the host to the target which is bound to be slow due to USB 2.0 FS bandwidth. It's unclear whether `probe-rs` exposes this operation in its API; it certainly implements the host to target `memcpy` operation.

- make it better:
  - a smarter way to do the 'paint the stack' part is to check if the stack was painted before (check lowest RAM address first? if it was, check for first non-occurrence of bit pattern) and then only paint the unpainted part

PR #254, opened 2021-09-06
Add support for measuring the program's stack usage

jonas-schievink:
This cleans up the stack canary code a bit, and adds a `--measure-stack` command that will paint the entire available stack region with a known bit pattern, and then report how much of the available stack space was used by the program.

Closes https://github.com/knurling-rs/probe-run/issues/188

```
(HOST) INFO  flashing program (7.03 KiB)
(HOST) INFO  success!
(HOST) INFO  painting 261063 bytes of RAM for stack usage estimation
0 INFO  recurse(times=10)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
1 INFO  recurse(times=9)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
2 INFO  recurse(times=8)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
3 INFO  recurse(times=7)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
4 INFO  recurse(times=6)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
5 INFO  recurse(times=5)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
6 INFO  recurse(times=4)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
7 INFO  recurse(times=3)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
8 INFO  recurse(times=2)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
9 INFO  recurse(times=1)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
10 INFO  recurse(times=0)
└─ overflow::recurse @ /home/jonas/dev/probe-test/src/bin/overflow.rs:13
(HOST) INFO  reading 261063 bytes of RAM for stack usage estimation
(HOST) INFO  program has used at least 23240/261063 bytes of stack space
(HOST) INFO  device halted without error
```

### DIFF
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -39,6 +39,7 @@ pub(crate) struct Canary {
     size: usize,
     stack_available: u32,
     data_below_stack: bool,
+    measure_stack: bool,
 }
 
 impl Canary {
@@ -46,6 +47,7 @@ impl Canary {
         sess: &mut Session,
         target_info: &TargetInfo,
         elf: &Elf,
+        measure_stack: bool,
     ) -> Result<Option<Self>, anyhow::Error> {
         let mut core = sess.core(0)?;
         core.reset_and_halt(TIMEOUT)?;
@@ -67,18 +69,28 @@ impl Canary {
 
         let stack_available = stack_range.end() - stack_range.start() - 1;
 
-        // We consider >90% stack usage a potential stack overflow, but don't go beyond 1 kb since
-        // filling a lot of RAM is slow (and 1 kb should be "good enough" for what we're doing).
-        let size = 1024.min(stack_available / 10) as usize;
+        let size = if measure_stack {
+            // When measuring stack consumption, we have to color the whole stack.
+            stack_available as usize
+        } else {
+            // We consider >90% stack usage a potential stack overflow, but don't go beyond 1 kb
+            // since filling a lot of RAM is slow (and 1 kb should be "good enough" for what we're
+            // doing).
+            1024.min(stack_available / 10) as usize
+        };
 
         log::debug!(
-            "{} bytes of stack available ({:#010X} ..= {:#010X}), using {} byte canary to detect overflows",
+            "{} bytes of stack available ({:#010X} ..= {:#010X}), using {} byte canary",
             stack_available,
             stack_range.start(),
             stack_range.end(),
             size,
         );
 
+        if measure_stack {
+            // Painting 100KB or more takes a few seconds, so provide user feedback.
+            log::info!("painting {} bytes of RAM for stack usage estimation", size);
+        }
         let address = *stack_range.start();
         let canary = vec![CANARY_VALUE; size];
         core.write_8(address, &canary)?;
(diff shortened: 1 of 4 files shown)


src/canary.rs
```diff
@@ -33,76 +33,127 @@ const CANARY_VALUE: u8 = 0xAA;
...
+                self.size
+            );
+        }
         let mut canary = vec![0; self.size];
         core.read_8(self.address, &mut canary)?;
 
-        if let Some(pos) = canary.iter().position(|b| *b != CANARY_VALUE) {
-            let touched_address = self.address + pos as u32;
-            log::debug!("canary was touched at {:#010X}", touched_address);
+        let min_stack_usage = match canary.iter().position(|b| *b != CANARY_VALUE) {
```
japaric, line 117: doesn't need to be addressed in this PR but this could be sped up (+) by doing a binary search in the `measure_stack` case. Though the above `read_8` will eat most of runtime so, instead of reading the whole `self.size` at once, this part should sample the stack at `sizeof(CANARY_VALUE)` chunks at a time and use that to perform a binary search.

(+) before doing any hypothetical perf improvement let's measure how long this takes vs how long the write takes. it may be that this is negligible compared to the write.

src/target_info.rs
```diff
@@ -1,29 +1,39 @@
...
             extract_active_ram_region(&probe_target, elf.vector_table.initial_stack_pointer);
-        let highest_static_var_address =
-            extract_highest_static_var_address(elf, active_ram_region.as_ref());
+        let stack_range = extract_stack_range(
+            elf,
+            active_ram_region.as_ref(),
+            elf.vector_table.initial_stack_pointer,
+        );
+
+        let data_below_stack = match (&active_ram_region, &stack_range) {
```
japaric, line 27: nit: this looks to me that the `TargetInfo` field should have type `Option<StackInfo>` and the `StackInfo` struct should be `{ stack_range: RangeInclusive<u32>, data_below_stack: bool }`

src/canary.rs
```diff
@@ -33,76 +33,127 @@ const CANARY_VALUE: u8 = 0xAA;
...
+            size,
+        );
+
+        if measure_stack {
+            // Painting 100KB or more takes a few seconds, so provide user feedback.
+            log::info!("painting {} bytes of RAM for stack usage estimation", size);
         }
+        let address = *stack_range.start();
+        let canary = vec![CANARY_VALUE; size];
+        core.write_8(address, &canary)?;
```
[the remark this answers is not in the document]
japaric: doesn't need to be done in this PR but it would be good add e.g. `trace` level logs that report the time it takes to write the canary, read the canary and write the program to flash. Just so that we have some measurements in place for when we look into optimizing things.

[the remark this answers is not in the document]
japaric: the other thing that I noticed was that number of bytes reported was an odd number. now I see that `CANARY_VALUE` is a byte. I was kind of expecting a word since the stack pointer is usually moved 4 or 8 bytes at a time. I don't think a 1-byte canary value is a problem right now given that we are scanning the stack linearly but if we switch to binary search it may result in false positives so the sampling done in the binary search should be several bytes (maybe 24 bytes at a time or so)

src/canary.rs
```diff
@@ -33,76 +33,127 @@ const CANARY_VALUE: u8 = 0xAA;
...
+            "{} bytes of stack available ({:#010X} ..= {:#010X}), using {} byte canary",
+            stack_available,
+            stack_range.start(),
+            stack_range.end(),
+            size,
+        );
+
+        if measure_stack {
+            // Painting 100KB or more takes a few seconds, so provide user feedback.
+            log::info!("painting {} bytes of RAM for stack usage estimation", size);
```
japaric, line 92: nit: could we report the number of bytes here in KB? either than use a comma to separate the thousands (e.g. `123,456`; see num-format crate). It's a bit hard to read at the moment will all the digits next to each other.

japaric (approved):
Looks great! I left some inline comments with nits around a data structure and the number representation.
Could you please open a few follow-up perf investigation issue? one for 'binary search read' and one for 'speed up canary write by having MCU do the memset'
Other than that, feel free to send this to bors :rocket:

src/canary.rs
jonas-schievink: I don't think the binary search optimization is going to work correctly, since the program won't touch *every* word in the used area of the stack, so the search might run into a `CANARY_VALUE` and incorrectly skip used areas of the stack

japaric: yes, I agree a 1-byte look-up is error prone hence binary search should read a bigger chunk at a time, see https://github.com/knurling-rs/probe-run/pull/254#discussion_r704169068 . Or maybe instead of considering a *single* partition point to decide if one half is used or not we could use 2 or more partition points ("add more depth to the binary partition tree"). Hard to explain without a diagram; will think more about it and file an issue if what I'm thinking of even makes sense.

PR #257, opened 2021-09-08
Split "set rtt to blocking-mode" into function

Urhengulas:
Firstly this shortens the `fn start_program`,
secondly it makes the "what" of this block of code more apparent.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -169,13 +169,13 @@ fn set_rtt_to_blocking(
     let rtt_buffer_address = rtt_buffer_address + OFFSET;
 
     // read flags
-    let rtt_control_block = &mut [0];
-    core.read_32(rtt_buffer_address, rtt_control_block)?;
+    let channel_flags = &mut [0];
+    core.read_32(rtt_buffer_address, channel_flags)?;
     // modify flags to blocking
     const BLOCK_IF_FULL: u32 = 2;
-    let b = rtt_control_block[0] | BLOCK_IF_FULL;
+    let modified_channel_flags = channel_flags[0] | BLOCK_IF_FULL;
     // write flags back
-    core.write_word_32(rtt_buffer_address, b)?;
+    core.write_word_32(rtt_buffer_address, modified_channel_flags)?;
 
     // clear the breakpoint we set before
     core.clear_hw_breakpoint(main_fn_address)?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -159,13 +159,25 @@ fn set_rtt_to_blocking(
     main_fn_address: u32,
     rtt_buffer_address: u32,
 ) -> anyhow::Result<()> {
+    // set and wait for a hardware breakpoint at the beginning of `fn main()`
     core.set_hw_breakpoint(main_fn_address)?;
     core.run()?;
     core.wait_for_core_halted(Duration::from_secs(5))?;
 
+    // calculate address of rtt control block
     const OFFSET: u32 = 44;
+    let rtt_buffer_address = rtt_buffer_address + OFFSET;
+
+    // read control block
+    let rtt_control_block = &mut [0];
+    core.read_32(rtt_buffer_address, rtt_control_block)?;
+    // modify it to blocking
     const BLOCK_IF_FULL: u32 = 2;
-    core.write_word_32(rtt_buffer_address + OFFSET, BLOCK_IF_FULL)?;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -160,12 +153,30 @@ fn start_program(sess: &mut Session, elf: &Elf) -> Result<(), anyhow::Error> {
...
+    core: &mut Core,
+    main_fn_address: u32,
+    rtt_buffer_address: u32,
+) -> anyhow::Result<()> {
+    core.set_hw_breakpoint(main_fn_address)?;
+    core.run()?;
+    core.wait_for_core_halted(Duration::from_secs(5))?;
+
+    const OFFSET: u32 = 44;
+    const FLAG: u32 = 2; // BLOCK_IF_FULL
```
Urhengulas, line 167: Would it make sense to import `BLOCK_IF_FULL` from `defmt-rtt`? Not that it would change the functionality, but more from a design perspective.

jonas-schievink: `defmt-rtt` will probably break on hosted systems, it targets microcontrollers and exports the RTT and defmt symbols.

Adding a dependency just for a single constant is also not really worth it IMO.

One thing we can do here is rename `FLAG` to `BLOCK_IF_FULL` instead of using a comment for that, that makes this a bit more descriptive.

src/main.rs
```diff
@@ -160,12 +153,30 @@ fn start_program(sess: &mut Session, elf: &Elf) -> Result<(), anyhow::Error> {
...
+    core.write_word_32(rtt_buffer_address + OFFSET, FLAG)?;
```
jonas-schievink, line 168: Preexisting issue, but it's odd that this just overwrites any existing flags. `defmt-rtt` initializes them with `NOBLOCK_TRIM`, but this will clear that flag. I would expect us to read-modify-write the flags instead.

Urhengulas: Sounds reasonable. How would I do that?

jonas-schievink: `read_word_32`, then `| FLAG`, then write that value back

src/main.rs
```diff
@@ -160,12 +153,42 @@ fn start_program(sess: &mut Session, elf: &Elf) -> Result<(), anyhow::Error> {
...
+    core: &mut Core,
+    main_fn_address: u32,
+    rtt_buffer_address: u32,
+) -> anyhow::Result<()> {
+    // set and wait for a hardware breakpoint at the beginning of `fn main()`
+    core.set_hw_breakpoint(main_fn_address)?;
+    core.run()?;
+    core.wait_for_core_halted(Duration::from_secs(5))?;
+
+    // calculate address of rtt control block
```
Urhengulas, line 167: I assumed that the word we are reading is the rtt control block and documented that in the comment. Is that correct?

Urhengulas: Is it better like it is now?

src/main.rs
```diff
@@ -160,12 +153,42 @@ fn start_program(sess: &mut Session, elf: &Elf) -> Result<(), anyhow::Error> {
...
+    rtt_buffer_address: u32,
+) -> anyhow::Result<()> {
+    // set and wait for a hardware breakpoint at the beginning of `fn main()`
+    core.set_hw_breakpoint(main_fn_address)?;
+    core.run()?;
+    core.wait_for_core_halted(Duration::from_secs(5))?;
+
+    // calculate address of up-channel-flags inside the rtt control block
+    const OFFSET: u32 = 44;
+    let rtt_buffer_address = rtt_buffer_address + OFFSET;
```
Urhengulas, line 169: Not saying that there would be a big gain, if at all, but from a "performance" perspective we probably should calculate the `rtt_buffer_address` in between `core.run` and `core.wait_for_core_halted`, shouldn't we? Since else we idle until we hit the breakpoint anyways.

jonas-schievink: No, it's the compiler's job to reorder code to make it faster

src/main.rs
```diff
@@ -160,12 +153,42 @@ fn start_program(sess: &mut Session, elf: &Elf) -> Result<(), anyhow::Error> {
...
+
+    // read flags
+    let rtt_control_block = &mut [0];
```
jonas-schievink, line 172: This should be renamed to something like `channel_flags` as it's not the control block

Urhengulas: Renamed it :)

PR #259, opened 2021-09-14
Update snapshot tests

japaric:
see individual commits for details

this makes `cargo test -- --ignored` (includes snapshot tests) pass again


japaric:
this particular log statement is somewhat problematic (but not the end of the world):
```
   19       │-(HOST) TRACE reading canary took 0.012s (83.93 KiB/s)
         19 │+(HOST) TRACE reading canary took 0.012s (81.14 KiB/s)
```
it prints the speed so it will make the snapshot test always fail (false positive). the snapshot test checks the output of probe-run when the  `--verbose` flag is passed

japaric:
CI failure is unrelated to this PR and should be fixed in PR #260

Issue #249, opened 2021-08-17
Make sure probe-run 0.3 can recover from decoding errors when framing is used


jonas-schievink:
Support for rzCOBS framing was added to the defmt decoder, which allows us to recover from malformed data instead of bailing out. probe-run needs to be capable of doing so, since otherwise framing is quite useless.

Companion issue to https://github.com/knurling-rs/defmt/issues/559

PR #264, opened 2021-09-27
use new stream decoder API

japaric:
fixes knurling-rs/defmt#575

marking as a draft because I want to update the manual snapshot tests before merging this


jonas-schievink:
I think it would be good to set up the unit tests to always be built from source, that makes it easier to modify them

bors r+

japaric:
> I think it would be good to set up the unit tests to always be built from source, that makes it easier to modify them

generally +1 to the idea of having the source in-tree. always rebuilding the snapshot binaries does have the complication that stuff can vary between systems and break the tests e.g. file paths, binary sizes, etc. resilience against that would require sort of wildcards in the .snap files (`flashing program ({digit*}.{digit*} KiB)`)) which I'm not sure insta supports?

src/main.rs
```diff
@@ -325,16 +330,12 @@ fn decode_and_print_defmt_logs(
...
-                let num_bytes = buffer.len();
-                buffer.rotate_left(consumed);
-                buffer.truncate(num_bytes - consumed);
             }
 
             Err(defmt_decoder::DecodeError::UnexpectedEof) => break,
 
             Err(defmt_decoder::DecodeError::Malformed) => {
-                log::error!("failed to decode defmt data: {:x?}", buffer);
+                log::error!("failed to decode defmt data");
```
japaric, line 338: by the way, @Urhengulas this looked like the right place to modify to implement #249 (I think)

Urhengulas: Thank you for the pointer!

PR #266, opened 2021-10-04
Recover from decoding-errors

### LINKED ISSUE #249: Make sure probe-run 0.3 can recover from decoding errors when framing is used

Urhengulas:
Fixes #249 

194a7f0db624afd901d46602aad41eb2e32fd8e1 is the important commit. The other two are drive-byes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,7 @@ use std::{
 
 use anyhow::{anyhow, bail};
 use colored::Colorize as _;
-use defmt_decoder::{Locations, StreamDecoder};
+use defmt_decoder::{DecodeError, Locations, StreamDecoder};
 use probe_rs::{
     flashing::{self, Format},
     Core,
@@ -254,6 +254,7 @@ fn extract_and_print_logs(
                             elf.defmt_locations.as_ref(),
                             current_dir,
                             opts,
+                            elf.defmt_table.as_ref().map(|t| t.encoding().can_recover()),
                         )?;
                     }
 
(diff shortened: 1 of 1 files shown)


Urhengulas:
Blocked by https://github.com/knurling-rs/probe-run/pull/267.

src/main.rs
```diff
@@ -298,52 +299,62 @@ fn decode_and_print_defmt_logs(
...
-
-            Err(defmt_decoder::DecodeError::Malformed) => {
-                log::error!("failed to decode defmt data");
-                return Err(defmt_decoder::DecodeError::Malformed.into());
-            }
+            Ok(frame) => forward_to_logger(&frame, locations, current_dir, shorten_paths),
+            Err(DecodeError::UnexpectedEof) => break,
+            Err(DecodeError::Malformed) => match encoding_can_recover {
+                // if recovery is impossible, or we don't know if it is, abort
+                Some(false) | None => return Err(DecodeError::Malformed.into()),
```
japaric, line 311: observation: the `None` case is unreachable because a `StreamDecoder` can only be constructed when an `ElfTable` exists and this encoding information can always be extracted from `ElfTable`

japaric (approved):
Looks good to me. If you want to address my observation (remove the `Option` around `encoding_can_recover`) go for it otherwise feel free to send this to bors as it is.

src/main.rs
japaric: I would suggest extracting the encoding info from [this line](https://github.com/knurling-rs/probe-run/blob/f03dcf195ca538dc9ec4afe0b6938e3017fcb181/src/main.rs#L226) so you would end up with a option of a tuple that contains the stream decoder and the encoding info.

Urhengulas: Kk. What about now?

PR #281, opened 2021-11-23
Report flashing size using probe-rs's FlashProgress system.

jonathanpallant:
We get an Initialized object at the start, and then one message per sector erase and one message per block write.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,10 +81,32 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     if opts.no_flash {
         log::info!("skipped flashing");
     } else {
-        let size = elf.program_flash_size();
-        log::info!("flashing program ({:.02} KiB)", size as f64 / 1024.0);
+        let fp = flashing::FlashProgress::new(|evt| {
+            match evt {
+                // The flash layout has been built and the flashing procedure was initialized.
+                flashing::ProgressEvent::Initialized { flash_layout, .. } => {
+                    let pages = flash_layout.pages();
+                    let num_pages = pages.len();
+                    let num_bytes: u64 = pages.iter().map(|x| x.size() as u64).sum();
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -81,10 +81,46 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     if opts.no_flash {
         log::info!("skipped flashing");
     } else {
-        let size = elf.program_flash_size();
-        log::info!("flashing program ({:.02} KiB)", size as f64 / 1024.0);
+        let fp = flashing::FlashProgress::new(|evt| {
+            match evt {
+                // The flash layout has been built and the flashing procedure was initialized.
+                flashing::ProgressEvent::Initialized { flash_layout, .. } => {
```
jonas-schievink, line 87: `FlashLayout` is not part of probe-rs' public API, are you sure we're supposed to use that?

jonathanpallant: In what sense is it not public? It's documented in https://docs.rs/probe-rs/0.11.0/probe_rs/flashing/enum.ProgressEvent.html#variant.Initialized.field.flash_layout as

> The layout of the flash contents as it will be used by the flash procedure. This is an exact report of what the flashing procedure will do during the flashing process.

It's also the only field in `ProgressEvent::Initialized` so it seems like it's OK to use it?

jonas-schievink: Yeah, but the type isn't documented, and you're calling methods on it which also aren't documented

jonathanpallant: I'll raise a ticket on probe-rs and check.

jonathanpallant: I spoke with @Dirbaio on Matrix, and basically it's now de-facto public API. They would like to replace it with something nicer (i.e. simpler) but that will be a breaking-change version bump, so we are safe to rely on this as long as we are on probe-rs 0.11.

Issue #255, opened 2021-09-07
Timeout on every second run


folkertdev:
**Describe the bug**
Every second run of `probe-run` results in a timeout

```
    Finished dev [optimized + debuginfo] target(s) in 0.02s
     Running `probe-run --chip STM32WLE5JBIx target/thumbv7em-none-eabi/debug/pingpong --probe 001A00195553500720393256`
(HOST) INFO  flashing program (18.83 KiB)
(HOST) INFO  success!
Error: Operation timed out
```
When there is no timeout, the program runs just fine. It is consistently every second time that a timeout occurs.

**To Reproduce**

For me, flash the supplied elf file to the device (in my case, an stm32wle). Every second time, this operation will timeout. The other time it will run just fine.

**config.toml**
The contents of your project's `.cargo/config.toml` file

``` toml
[target.'cfg(all(target_arch = "arm", target_os = "none"))']
runner = "probe-run --chip STM32WLE5JBIx"

rustflags = [
  "-C", "linker=flip-link",
  "-C", "link-arg=-Tlink.x",
  "-C", "link-arg=-Tdefmt.x",
  # This is needed if your flash or ram addresses are not aligned to 0x10000 in memory.x
  # See https://github.com/rust-embedded/cortex-m-quickstart/pull/95
  "-C", "link-arg=--nmagic",
]

[build]
target = "thumbv7em-none-eabi"   # Cortex-M4 and Cortex-M7 (no FPU)

[alias]
rb = "run --bin"
rrb = "run --release --bin"

```

**Probe details**
You can get this information from [`probe-rs-cli`](https://crates.io/crates/probe-rs-cli). Your microcontroller must be connected to your PC / laptop when you run the command below.

*Example:*

``` console
[0]: STLink V3 (VID: 0483, PID: 374e, Serial: 001A00195553500720393256, STLink)
```

**Operating System:**
Linux (mint 19)

**ELF file (attachment)**

[pingpong.tar.gz](https://github.com/knurling-rs/probe-run/files/7121583/pingpong.tar.gz)

**Additional context**

@jamesmunns  suggested that configurable timeouts/retries may be a solution here.

jonas-schievink:
Does this also happen with `cargo embed`?

folkertdev:
I'm not sure I understand. The flashing process always succeeds. Can `cargo embed` actually (attempt to) run the application? It's not clear how to do that from the `--help` message or the README

Urhengulas:
[...]

PR #285, opened 2021-11-25
Update probe-rs, probe-rs-rtt to 0.12

korken89:
Probe-rs 0.12 was just released. This updates the dependency and fixes the breaking changes.

Tested the binary using an STM32 Nucleo devboard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
-- [#]
+- [#285]: Update `probe-rs` and `probe-rs-rtt` to `0.12`
 
+[#285]: https://github.com/knurling-rs/probe-run/pull/285
 
 ## [v0.3.0] - 2021-11-09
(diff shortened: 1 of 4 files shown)


jonathanpallant (approved):

jonathanpallant:
Thanks for this @korken89! I'm very excited to see RP2040 support :)

Works OK for me on an STM32 Nucleo as well. Would like to hear from @japaric or @jonas-schievink or someone with an nRF52840 just as a double-check before merging though.

jonathanpallant:
Edit, works OK on the Raspberry Pi Pico too (using a second Pico running [the CMSIS-DAP firmware](https://github.com/majbthrd/DapperMime)).

newAM:
This fixes #255

jonas-schievink:
A quick test on the nRF52840 seems to work fine as well

newAM:
Would it be possible to get a release with this change included?  This would help out a lot of people using the STM32WL; it fixes #255 and https://github.com/stm32-rs/stm32wlxx-hal/issues/74.

Issue #277, opened 2021-11-15
False positive "call stack was corrupted" warning on 0.3


jonas-schievink:
**Describe the bug**
I'm seeing a `(HOST) WARN  call stack was corrupted; unwinding could not be completed`, even though unwinding seems to have worked fine and nothing happened that could have corrupted the stack.

**To Reproduce**

- Check out https://github.com/stm32-rs/bxcan/pull/47
- `cd testsuite; cargo test`

**Expected and observed behavior**

Actual output:

```
    Finished test [unoptimized + debuginfo] target(s) in 0.01s
     Running tests/integration.rs (/home/jonas/dev/bxcan/target/thumbv7m-none-eabi/debug/deps/integration-7c7caa16bcb9d813)
(HOST) INFO  flashing program (16.20 KiB)
(HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
(1/11) running `split_filters`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:18
(2/11) running `basic_roundtrip`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:43
(3/11) running `no_filters_no_frames`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:58
(4/11) running `filter_mask32_std`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:68
(5/11) running `filter_mask32_ext`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:112
(6/11) running `filter_mask16`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:153
(7/11) running `filter_list32_std`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:193
(8/11) running `filter_list32_ext`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:226
(9/11) running `dequeue_lower_priority_frame`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:264
(10/11) running `enable_non_blocking`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:317
(11/11) running `ext_roundtrip`...
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:330
all tests passed!
└─ integration::tests::__defmt_test_entry @ tests/integration.rs:4
────────────────────────────────────────────────────────────────────────────────
stack backtrace:
[...]

PR #287, opened 2021-11-25
unwind: skip FDEs with initial address of 0

jonas-schievink:
The [`panic` example in the app-template](https://github.com/knurling-rs/app-template/blob/09cb051ef0e6863e805a3fc65d509147b031f411/src/bin/panic.rs) can cause a broken backtrace like this:

```
stack backtrace:
   0: 0x00001450 @ HardFaultTrampoline
      <exception entry>
   1: 0x000008d4 @ lib::inline::__udf
        at ./asm/inline.rs:172:5
   2: 0x000008d4 @ __udf
        at ./asm/lib.rs:49:17
   3: 0x000001ba @ cortex_m::asm::udf
        at [...]/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.7.3/src/asm.rs:43:5
   4: 0x000001c4 @ _defmt_panic
        at [...]/src/lib.rs:13:5
   5: 0x000001b0 @ defmt::export::panic
        at [...]/.cargo/registry/src/github.com-1ecc6299db9ec823/defmt-0.3.0/src/export/mod.rs:125:14
   6: 0x000001b0 @ panic::__cortex_m_rt_main
        at [...]/src/bin/panic.rs:10:5
   7: 0x00000168 @ main
        at [...]/src/bin/panic.rs:6:1
   8: 0x00000144 @ Reset
   9: 0x00000100 @ Reset
  10: 0x52205244 @ __sheap
```

This appeared to have a different cause from https://github.com/knurling-rs/probe-run/issues/277, since the fix for that didn't fix the backtrace here.

After a lot of investigation, it turned out that what was going on is that there are multiple frame description entries (FDEs) in the debuginfo that claim to apply to some addresses, and we happened to pick the wrong one while unwinding from `Reset` in frame 8, causing us to restore registers incorrectly and then attempting to jump to a nonsense address.

This seems to be limited to FDEs describing presumably dead code, whose base address gets reset to 0 when they're included in the final ELF, so this PR fixes it by ignoring any FDEs that claim to apply to address 0.

### DIFF
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -1,7 +1,10 @@
 //! unwind target's program
 
 use anyhow::{anyhow, Context as _};
-use gimli::{BaseAddresses, DebugFrame, UninitializedUnwindContext, UnwindSection as _};
+use gimli::{
+    BaseAddresses, CieOrFde, DebugFrame, FrameDescriptionEntry, Reader, UninitializedUnwindContext,
+    UnwindSection as _,
+};
 use probe_rs::{config::RamRegion, Core};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


jonathanpallant (approved):
unwind_info_for_address being a method on both UnwindSection and FrameDescriptionEntry threw me for a loop there ... but this LGTM now I understand what's going on in the diff!

PR #295, opened 2022-01-12
`probe-run` json output

Urhengulas:
This PR adds the `--json` flag, which makes use of the new `JsonLogger` which is getting added to `defmt-decoder` in https://github.com/knurling-rs/defmt/pull/651.


Urhengulas:
bors retry

PR #299, opened 2022-02-02
Fix and refactor `fn extract_stack_info`

Urhengulas:
- Fix the `stack_range` end.
    The high end of the stack is always `initial_stack_pointer - 1`.
- Use range methods instead of variables.
- Simplify accessing range of ram region
- Introduce `fn is_superset` for two ranges.
    I am aware that in our concrete scenario we don't need all four, but
    only two conditions (as it was before), since we are checking
    `subset.contains(superset.end())` before. But I have added them for
    completeness.
- Document what `fn extract_stack_info` does
- Simplify arguments of `fn extract_stack_info`

### DIFF
--- a/src/target_info.rs
+++ b/src/target_info.rs
@@ -94,9 +94,7 @@ fn extract_stack_info(
                     name
                 );
                 return None;
-            } else if stack_range.end() >= section_range.start()
-                && stack_range.start() <= section_range.end()
-            {
+            } else if is_superset(&stack_range, &section_range) {
                 stack_range = section_range.end() + 1..=*stack_range.end();
             }
         }
@@ -110,4 +108,34 @@ fn extract_stack_info(
         data_below_stack: *stack_range.start() > ram_range.start,
         range: stack_range,
     })
+}
+
+fn is_superset(superset: &RangeInclusive<u32>, subset: &RangeInclusive<u32>) -> bool {
+    subset.start() >= superset.start()
+        && subset.start() <= superset.end()
+        && subset.end() <= superset.end()
+        && subset.end() >= superset.start()
+}
+
+#[cfg(test)]
+mod tests {
+    use rstest::rstest;
+
+    use super::*;
+
+    #[rstest]
+    #[case(0..=10, 0..=10, true)]
+    #[case(0..=10, 1..=9, true)]
+    #[case(0..=10, 0..=5, true)]
+    #[case(0..=10, 5..=10, true)]
+    #[case(0..=10, 0..=11, false)]
+    #[case(0..=10, 5..=11, false)]
+    fn should_extract_hash_from_description(
+        #[case] superset: RangeInclusive<u32>,
+        #[case] subset: RangeInclusive<u32>,
+        #[case] expected: bool,
+    ) {
+        let is_superset = is_superset(&superset, &subset);
+        assert_eq!(is_superset, expected)
+    }
 }
--- a/src/target_info.rs
+++ b/src/target_info.rs
@@ -97,7 +97,7 @@ fn extract_stack_info(elf: &Elf, ram_range: &Range<u32>) -> Option<StackInfo> {
                     name
                 );
                 return None;
-            } else if is_superset(&(stack_start..=stack_end), &section_range) {
+            } else if section_range.contains(&stack_start) {
                 stack_start = section_range.end() + 1;
             }
         }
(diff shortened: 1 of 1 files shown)


Urhengulas:
If my understanding of `fn extract_stack_info` is correct we could simplify it even more, by just selecting the `highest_address` for each non-empty section that lies inside RAM and then taking the maximum of that plus one as the lower end of the stack.

What do you think?

---

_It would look like this:_
```rust
fn extract_stack_info(elf: &Elf, ram_region: Option<&RamRegion>) -> Option<StackInfo> {
    // How does it work?
    // - the upper end of the stack is the initial SP, minus one
    // - the lower end of the stack is the highest address any section in the elf file uses, plus one

    let ram_range = &ram_region?.range;
    let highest_section = elf
        .sections()
        .filter_map(|section| {
            let size: u32 = section.size().try_into().expect("expected 32-bit ELF");
            if size == 0 {
                return None;
            }

            let lowest_address: u32 = section.address().try_into().expect("expected 32-bit ELF");
            let highest_address = lowest_address + size - 1;
            let name = section.name().unwrap_or("<unknown>");

            if ram_range.contains(&highest_address) {
                log::debug!(
                    "section `{}` is in RAM at {:#010X?}",
                    name,
                    lowest_address..=highest_address
                );
                Some(highest_address)
            } else {
                None
            }
        })
        .max()?;

    let initial_stack_pointer = elf.vector_table.initial_stack_pointer;
    let stack_range = highest_section + 1..=initial_stack_pointer - 1;
    log::debug!("valid SP range: {:#010X?}", stack_range);
    Some(StackInfo {
        data_below_stack: *stack_range.start() > ram_range.start,
        range: stack_range,
    })
}
```

src/target_info.rs
```diff
@@ -85,34 +83,50 @@ fn extract_stack_info(
...
-                "section `{}` is in RAM at {:#010X} ..= {:#010X}",
-                name,
-                lowest_address,
-                highest_address,
-            );
+        if ram_range.contains(section_range.end()) {
+            log::debug!("section `{}` is in RAM at {:#010X?}", name, section_range);
 
-            if section_range.contains(&(initial_stack_pointer - 1)) {
+            if section_range.contains(stack_range.end()) {
```
jonas-schievink, line 89: This changes behavior and makes the logged message below inaccurate

jonas-schievink: Ah, nevermind, `stack_range.end()` is always constant. Perhaps we should just use `let stack_end = initial_stack_pointer - 1;` and only construct the range when returning, that sounds clearer.

Urhengulas: 👍🏾

Urhengulas: It actually took me quite a while to understand the function, part of it because there were multiple variables representing pretty much the same thing, so I also tried to unify that by only using the ranges.

src/target_info.rs
```diff
@@ -64,15 +61,17 @@ fn extract_active_ram_region(
...
-    active_ram_region: Option<&RamRegion>,
-    initial_stack_pointer: u32,
-) -> Option<StackInfo> {
-    let active_ram_region = active_ram_region?;
+fn extract_stack_info(elf: &Elf, ram_region: Option<&RamRegion>) -> Option<StackInfo> {
+    // How does it work?
+    // - the upper end of the stack is the initial SP, minus one
+    // - the lower end of the stack is the highest address any section in the elf file uses, plus one
+
+    let ram_range = &ram_region?.range;
```
jonas-schievink, line 69: If you want to do some more cleanup: `ram_region?` could be moved out of the function, there isn't really a reason to accept an `Option` here, but then immediately early-return when it's `None`.

src/target_info.rs
```diff
@@ -85,34 +84,52 @@ fn extract_stack_info(
...
                 log::debug!(
                     "initial SP is in section `{}`, cannot determine valid stack range",
                     name
                 );
                 return None;
-            }
-
-            if initial_stack_pointer > lowest_address && *stack_range.start() <= highest_address {
-                stack_range = highest_address + 1..=initial_stack_pointer;
+            } else if is_superset(&(stack_start..=stack_end), &section_range) {
```
jonas-schievink, line 96: I think this doesn't match what the original code was trying to do. We don't really care about whether one range contains the other, we want to know if the current `stack_start` needs to be adjusted because of `section_range`, so the code should check if `stack_start` is in the section range, and adjust that accordingly.

Urhengulas: > so the code should check if `stack_start` is in the section range

Don't we just need to check if the current `stack_start` is below `section_range.end()`?

jonas-schievink: Ah, yeah, it doesn't have to be *inside* the section

Urhengulas:
@jonas-schievink The PR should be ready now. I tested it with the app-template and nrf52840-dk.

Issue #291, opened 2021-12-07
probe-rs does not support passing configuration for "--chip-description-path" argument to underlying probe-rs


TomDeRybel:
**Describe the bug**
I'm using parts in the STM32H735 family, which recently gained PAC and HAL support. Unfortunately these parts aren't supported by probe-rs yet (no yaml file entries for them in the crate). Thus, I used the procedure indicated by probe-rs to generate a local yaml from the pack file provided by ARM. This seems to all work fine using the probe-rs tools, such as cargo-flash, which support the use of the "--chip-description-path" argument.

To allow probe-run to do the same, I would need to be able to configure the path to this self-generated yaml, which currently does not seem supported.

**Expected and observed behavior**
I expect an argument to specify the path to the underlying probe-rs in .cargo/config, for instance:
runner = "probe-run --chip STM32H723VGTx --chip-description-path ../probe-rs_generated_packs/STM32H7_Series.yaml"

**Operating System:**
Linux

jonas-schievink:
This seems like something that would potentially be better solved with an environment variable that adds search paths for chip descriptions (that way the change applies across tools)

TomDeRybel:
An environment variable would work for me.
At this time, the probe-rs tools use different mechanisms themselves: 
- cargo-flash uses a command-line parameter 
- cargo-embed uses a definition file (Embed.toml)

TomDeRybel:
Thank you for adding this feature!

Urhengulas:
> Thank you for adding this feature!

You should thank @KOBA789 for implementing it! 👏🏾

TomDeRybel:
Thank you @KOBA789! This helps a lot with my own STM32H735 adventures :champagne:

PR #301, opened 2022-02-21
Add way to pass chip description file

KOBA789:
Fixes: https://github.com/knurling-rs/probe-run/issues/291

I need the way to pass my own chip description file to probe-run because my target chip is a original soft processor core implemented on a FPGA.

There are other ideas on #291 but I wrote the simplest implementation for now.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -27,7 +27,8 @@ pub(crate) struct Opts {
     #[structopt(long, required_unless_one(&["list-chips", "list-probes", "version"]), env = "PROBE_RUN_CHIP")]
     chip: Option<String>,
 
-    #[structopt(name = "chip description file path", long = "chip-description-path")]
+    /// Path to chip description file, in YAML format.
+    #[structopt(long)]
     pub(crate) chip_description_path: Option<PathBuf>,
 
     /// The probe to use (eg. `VID:PID`, `VID:PID:Serial`, or just `Serial`).
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -27,6 +27,9 @@ pub(crate) struct Opts {
     #[structopt(long, required_unless_one(&["list-chips", "list-probes", "version"]), env = "PROBE_RUN_CHIP")]
     chip: Option<String>,
 
+    #[structopt(name = "chip description file path", long = "chip-description-path")]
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -27,6 +27,9 @@ pub(crate) struct Opts {
     #[structopt(long, required_unless_one(&["list-chips", "list-probes", "version"]), env = "PROBE_RUN_CHIP")]
     chip: Option<String>,
 
+    #[structopt(name = "chip description file path", long = "chip-description-path")]
+    pub(crate) chip_description_path: Option<PathBuf>,
```
Urhengulas, line 31:
```suggestion
    /// Path to chip description file, in YAML format.
    #[structopt(long)]
    pub(crate) chip_description_path: Option<PathBuf>,
```

Urhengulas (changes requested):
Thank you for the PR @KOBA789!

When specifying the `--chip-description-path`, does a user still need to specify the `--chip` additionally? I don't ask what `clap` requires, but `probe-rs`.

src/main.rs
```diff
@@ -52,6 +52,9 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let elf_bytes = fs::read(elf_path)?;
     let elf = &Elf::parse(&elf_bytes)?;
 
+    if let Some(ref cdp) = opts.chip_description_path {
+        probe_rs::config::add_target_from_yaml(Path::new(cdp))?;
+    }
```
Urhengulas, line 57:
```suggestion
    if let Some(cdp) = &opts.chip_description_path {
        probe_rs::config::add_target_from_yaml(Path::new(cdp))?;
    }
```

KOBA789:
Thank you for your review.

> When specifying the --chip-description-path, does a user still need to specify the --chip additionally? I don't ask what clap requires, but probe-rs.

A chip description file may contains some variants. I think `--chip` is necsessary to specify which the variant to use.

Urhengulas:
> A chip description file may contains some variants. I think `--chip` is necsessary to specify which the variant to use.

Makes sense, I just wanted to check with you.

bors r+

PR #302, opened 2022-02-21
Make stack painting fast again! 🇪🇺 

### LINKED ISSUE #258: Stack painting with `--measure-stack` is slow
With `--measure-stack`, added in https://github.com/knurling-rs/probe-run/pull/254, we paint the whole area the stack could occupy with a bit pattern, and then read it back to determine the program's stack usage. This can write and read hundreds of KBs of RAM, which takes several seconds, so it would be great to speed this up.

One idea for speeding this up was to essentially run `memset` on the MCU, but probe-rs does not seem to expose an API for this (if this is even possible at all, with the vendor-provided on-device algorithms).

Urhengulas:
This PR implements the first one of improvements outlined in #258.

Fixes #258.

## But what is "stack painting" anyways?

The idea is to write a specific byte pattern to (part of) the stack before the program is getting executed. After the program finished, either because it is done with its task, or because there was an error, we read out the previously painted area and check how much of it is still intact. If the pattern is still the same, we can be rather certain that the program didn't write to this part of the stack. This information helps to either know if there was a stack overflow, or just to measure how much of the stack was used.

So far both reading and writing of the memory was done via the probe. While this works it is also rather slow, because the host and probe communicate via USB which takes time.

The new approach is writing a subroutine to the MCU, which will paint the memory from within.

## Mesurements

In following table you can see the measurement how much time the old and new approach take for memory from 8 to 256KiB.

data

The results are pretty impressive. The new approach is about 170 times faster!

## Further work

- A similar approach can also be applied to reading out the stack after the program finished.
- Additionally the stack canary can be simplified quite a lot. So far we are not painting the whole stack, except the user asks for it, because this _was_ slow. Because it is fast now we can always paint all of it, which simplifies the code and removes the need for the `--measure-stack` flag.

### DIFF
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -258,16 +258,19 @@ fn subroutine(start: u32, end: u32) -> [u8; SUBROUTINE_LENGTH] {
 
     const CV: u8 = CANARY_VALUE;
     [
-        0x03, 0x48, // ldr
-        0x04, 0x49, // ldr
-        0x04, 0x4a, // ldr
-        0x81, 0x42, // cmp
-        0x01, 0xD0, // beq.n
-        0x04, 0xC0, // stmia
-        0xFB, 0xE7, // b.n
-        0x00, 0xBE, // bkpt
-        s1, s2, s3, s4, // start address
-        e1, e2, e3, e4, // end address
-        CV, CV, CV, CV, // canary value
+        0x03, 0x48, // ldr r0, [pc, #12]
+        0x04, 0x49, // ldr r1, [pc, #16]
+        0x04, 0x4a, // ldr r2, [pc, #16]
+        // <loop>
+        0x81, 0x42, // cmp r1, r0
+        0x01, 0xD0, // beq.n   116 <end>
+        0x04, 0xC0, // stmia   r0!, {r2}
+        0xFB, 0xE7, // b.n 10e <loop>
+        // <end>
+        0x00, 0xBE, // bkpt    0x0000
+        //
+        s1, s2, s3, s4, // .word ; start address
+        e1, e2, e3, e4, // .word ; end address
+        CV, CV, CV, CV, // .word ; canary value
     ]
 }
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -193,18 +193,16 @@ fn paint_stack(core: &mut Core, start: u32, mut end: u32) -> Result<(), probe_rs
         end += end % 4;
     }
 
-    // construct subroutine
-    let subroutine = subroutine(start + SUBROUTINE_LENGTH, end);
-
     // does the subroutine fit inside the stack?
     let stack_size = (end - start) as usize;
     assert!(
-        subroutine.len() < stack_size,
+        SUBROUTINE_LENGTH < stack_size,
         "subroutine doesn't fit inside stack"
     );
 
     // write subroutine to RAM
-    core.write_8(start, &subroutine)?;
+    // NOTE: add `SUBROUTINE_LENGTH` to `start`, to avoid the subroutine overwriting itself
+    core.write_8(start, &subroutine(start + SUBROUTINE_LENGTH as u32, end))?;
 
     // store current PC and set PC to beginning of subroutine
     let previous_pc = core.read_core_reg(PC)?;
@@ -215,7 +213,7 @@ fn paint_stack(core: &mut Core, start: u32, mut end: u32) -> Result<(), probe_rs
     core.wait_for_core_halted(TIMEOUT).unwrap();
 
     // overwrite subroutine
-    core.write_8(start, &vec![CANARY_VALUE; subroutine.len()])?;
+    core.write_8(start, &[CANARY_VALUE; SUBROUTINE_LENGTH])?;
 
     // reset PC to where it was before
     core.write_core_reg(PC, previous_pc)?;
@@ -224,9 +222,7 @@ fn paint_stack(core: &mut Core, start: u32, mut end: u32) -> Result<(), probe_rs
 }
 
 /// The length of the subroutine.
-///
-/// This is checked on runtime and needs to be updated when changing the subroutine.
-const SUBROUTINE_LENGTH: u32 = 28;
+const SUBROUTINE_LENGTH: usize = 28;
 
 /// Create a subroutine to paint [`CANARY_VALUE`] from `start` till `end`.
 ///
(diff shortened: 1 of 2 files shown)


src/canary.rs
```diff
@@ -186,3 +185,88 @@ impl Canary {
         }
     }
 }
+
+/// Write [`CANARY_VALUE`] to the stack.
+fn paint_stack(core: &mut Core, start: u32, mut end: u32) -> Result<(), probe_rs::Error> {
```
jonathanpallant, line 190: I'd add a note here to observe that `start` should be the numerically lower address, and `end` should be the numerically higher address: that is `start < end`.

jonathanpallant: Maybe also panic if that is not upheld. Otherwise the subtraction at line 197 will underflow and panic.

jonathanpallant: Does `start` also have to be aligned?

Urhengulas: > I'd add a note here to observe that `start` should be the numerically lower address, and `end` should be the numerically higher address: that is `start < end`.

Good point, I am asserting this in line 250 in `fn subroutine`, but it is probably better to already check it before.



> Does `start` also have to be aligned?

I think so, but I was not sure how to go about it if it is not. Probably subtracting the modulo, right?

jonathanpallant: > Good point, I am asserting this in line 250 in fn subroutine, but it is probably better to already check it before.

I'd apply the comment to whichever functions take `start` and `end` as arguments :)

jonathanpallant: > Probably subtracting the modulo, right?

Yes, I think so. Or just bailing out - the linker script really should have set the stack start and end to be 4-byte aligned anyway, plus https://developer.arm.com/documentation/ddi0403/d/System-Level-Architecture/System-Level-Programmers--Model/ARMv7-M-exception-model/Stack-alignment-on-exception-entry says:


> The ARMv7-M architecture guarantees that stack pointer values are at least 4-byte aligned.

src/canary.rs
```diff
@@ -186,3 +185,88 @@ impl Canary {
...
+    // does the subroutine fit inside the stack?
+    let stack_size = (end - start) as usize;
+    assert!(
+        SUBROUTINE_LENGTH < stack_size,
+        "subroutine doesn't fit inside stack"
+    );
+
+    // write subroutine to RAM
+    // NOTE: add `SUBROUTINE_LENGTH` to `start`, to avoid the subroutine overwriting itself
+    core.write_8(start, &subroutine(start + SUBROUTINE_LENGTH as u32, end))?;
```
jonathanpallant, line 205: If we're 4-byte aligned, I guess we could use `write_32` here? Not sure it makes much difference though.

src/canary.rs
```diff
@@ -186,3 +185,88 @@ impl Canary {
...
+    assert_eq!(start % 4, 0, "`start` needs to be 4-byte-aligned");
+    assert_eq!(end % 4, 0, "`end` needs to be 4-byte-aligned");
+
+    // convert start and end address to bytes
+    let [s1, s2, s3, s4] = start.to_le_bytes();
+    let [e1, e2, e3, e4] = end.to_le_bytes();
+
+    const CV: u8 = CANARY_VALUE;
+    [
+        0x03, 0x48, // ldr
```
jonathanpallant, line 260: I'd paste the full instruction in the comment, so `ldd r0, [pc, #12]`.

jonathanpallant (approved):
I had some minor suggestions for improving comments (specifically around document pre-conditions like address alignment). But otherwise this is brilliant! Great work :) 👍

src/canary.rs
Urhengulas: Do you think we need then need it also above the function? Or is inside enough?

jonathanpallant: I like the one above as it is because it's a full objdump. But in this arrays there are three LDR instructions in a row, and I think we just need a little bit more context to see which one is which.

src/canary.rs
Urhengulas: I don't think this works so well. It would require us to make `fn subroutine` return a `[u32; _]`, which would require is to combine two instructions (4 bytes) into one, which makes it less readable imo.

Or what do you think?

```diff
-    [
-        0x03, 0x48, // ldr
-        0x04, 0x49, // ldr
-        0x04, 0x4a, // ldr
-        0x81, 0x42, // cmp
-        0x01, 0xD0, // beq.n
-        0x04, 0xC0, // stmia
-        0xFB, 0xE7, // b.n
-        0x00, 0xBE, // bkpt
-        s1, s2, s3, s4, // start address
-        e1, e2, e3, e4, // end address
-        CV, CV, CV, CV, // canary value
-    ]
+    [
+        0x03480449, // ldr, ldr
+        0x044A8142, // ldr, cmp
+        0x01D004C0, // beq.n, stmia
+        0xFBE700BE, // b.n, bkpt
+        start, // start address
+        end, // end address
+        canary, // canary value
+    ]
```

src/canary.rs
Urhengulas: Interesting. With the nrf52840-dk and the `nrf52840-hal` I observe the `end` to be not 4-, but only 2-byte-aligned.

Before aligning it:
```text
start: 0x20000000, start % 4 == 0
end: 0x2003FBC6, end % 4 == 2
```
After aligning it: 
```
start: 0x20000000, start % 4 == 0
end: 0x2003FBC8, end % 4 == 0
```

Urhengulas: The value of `end` should be the initial stack pointer minus one.

Urhengulas: My only concern is, since the initial stack pointer points to the address directly _above_ the stack, it itself doesn't belong to the stack anymore. Can anything bad happen when we are writing to this address? (I don't think so because static variables and so on will only be created later, but I still want to double check).

Urhengulas:
The error messages can probably also be improved, but I'd like to do this in a follow-up PR, when reworking the canary (see "Additionally the stack canary can be simplified quite a lot. [...]".

src/canary.rs
jonathanpallant: If the stack is at the top of the RAM, the initial stack pointer value will be an invalid address.

From `memory.x` in `cortex-m-rt`:

> `_stack_start = ORIGIN(CCRAM) + LENGTH(CCRAM);`

The problem, I think, is a range doing `-1` when we should be doing `-4`, because we work in 4-byte units. Or perhaps we shouldn't hold *start* and *end* and instead hold *start* and *length-in-32-bit-words*.

Urhengulas: > The problem, I think, is a range doing `-1` when we should be doing `-4`, because we work in 4-byte units. Or perhaps we shouldn't hold *start* and *end* and instead hold *start* and *length-in-32-bit-words*.

So changing it to `initial_stack_pointer - 4` should be fine? 
I'd like to avoid bigger refactorings of other parts of the code as part of this PR.

Urhengulas: See https://github.com/knurling-rs/probe-run/pull/302/commits/67c15760a1eeb13f006ec371562917c36c16dfa5.

Urhengulas:
bors cancel

Urhengulas:
bors r=jonathanpallant

PR #305, opened 2022-03-01
Add option to disable double buffering while loading firmware

berkowski:
probe-rs/probe-rs#1030 exposes an option to disable double-buffering when downloading firmware, which seems required currently for the STM32L071x family of chips (probe-rs/probe-rs#883).  This PR allows use of that option for `probe-run`

This won't compile now as Cargo.toml is using the released 0.12 verson of probe-rs.  I've tested it locally by adding the following patch:

```toml
# Cargo.toml
[patch."crates-io"]
probe-rs = {git = "https://github.com/probe-rs/probe-rs", version = "0.12"}
probe-rs-rtt = {git = "https://github.com/probe-rs/probe-rs", version = "0.12"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - [#319] Warn on target chip mismatch
 - [#317] Clarify "can't determine stack overflow" error message
 - [#314] Clarify documentation in README
+- [#305] Add option to disable double buffering while loading firmware
 - [#293] Update snapshot tests to new TRACE output
 
 [#339]: https://github.com/knurling-rs/probe-run/pull/339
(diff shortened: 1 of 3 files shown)


japaric:
thanks for the PR. evaluating this will need to wait until there is a new probe-rs release on crates.io (0.13.0 I guess as there are breaking changes in the API).

berkowski:
No problem on the delay -- mostly wanted to get it up here so I wouldn't forget myself and maybe it helps someone else.  Still working fine for me with a local build and the [patch] section.

Thanks for all the effort you've been putting into rust on stm32.   It's been interesting to watch it over the last couple of years.

Urhengulas:
@berkowski I've rebased the PR. Are we good to merge?

berkowski:
Seems to work for me.  I successfully flashed my STM32L07x based project using this rebased PR and the --disable-double-buffering argument.  Defmt logs come through fine.

Urhengulas:
It also works for me on the nrf52840-dk. Thank you for your effort! 👍🏾 

bors r+

berkowski:
Awesome, looking forward to the capability making it into a release!

PR #318, opened 2022-04-06
Disable output colorization on terminals that don't support it

fhars:
The current version of the colored crate ignores the capabilities of
the terminal (see https://github.com/mackwic/colored/issues/108),
which makes the outupt of probe-run unparseable in e.g. emacs
compilation mode. As a temporary workaround, this patch overrides
colorization if probe-run is called from a terminal without color
support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ probe-rs = "0.12"
 probe-rs-rtt = "0.12"
 signal-hook = "0.3"
 structopt = "0.3"
+terminfo = "0.7"
 
 [dev-dependencies]
 dirs = "4"
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,18 @@ use crate::{backtrace::Outcome, canary::Canary, elf::Elf, target_info::TargetInf
 const TIMEOUT: Duration = Duration::from_secs(1);
 
 fn main() -> anyhow::Result<()> {
+    // colored 2.0.0 ignores the capabilities of the terminal, so we
+    // disable colorization here for terminals that don't support it
+    // until https://github.com/mackwic/colored/issues/108 is fixed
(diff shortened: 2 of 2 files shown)


Urhengulas (changes requested):
Hi @fhars, thank you for the patch. I tested it and it seems to work as expected. Can you please add the `Cargo.lock` file as well?

I have two concerns regarding the added `terminfo` dependency:
1. The maintenance of it seems to be currently unclear. See https://github.com/meh/rust-terminfo/issues/20. There are a few people interested in taking over maintenance, but it seems the handover didn't happen yet.
2. It adds outdated versions for a few dependencies, which some of them are already part of our dependency tree. This partially like a symptom of point 1.

*I think* it is reasonable to merge this PR anyways, but we should keep an eye on that and if it stays unmaintained eventually look for an alternative crate.

jonas-schievink:
Does emacs' compilation mode present itself as a terminal (via `isatty`) to the program (I do hope we support non-isatty mode properly already, so perhaps it doesn't)? Does it set something like `TERM=dumb` that we could detect instead?

Urhengulas:
> Does emacs' compilation mode present itself as a terminal (via `isatty`) to the program (I do hope we support non-isatty mode properly already, so perhaps it doesn't)? Does it set something like `TERM=dumb` that we could detect instead?

From reading https://github.com/mackwic/colored/issues/108 I understood that it is setting `TERM=dumb`. But I don't know if this is the only trigger we want to look out for, or if there is other ones as well?

Urhengulas:
@fhars Since it is a temporary fix until `colored` handles it properly, would it be enough for your purposes to only check `TERM=dumb` and disable coloring then?

Urhengulas:
Superseded by #320

PR #319, opened 2022-04-14
Warn on target chip mismatch

### LINKED ISSUE #115: Check the selected processor and target for compatability.
After spending an inordinate amount of time trying to work out why RTT wasn't working, I discovered that I had set the wrong target.

As ```probe-run``` knows both the source target (through the path) and the target IC, it should be possible to show a warning if the two aren't compatible with each other.

Urhengulas:
This PR makes `probe-run` emit a warning, in case the chip type (obtained from `probe-rs`) and the compilation target (obtained from the path to the elf file) do not match. The path will contain the compilation target if `probe-run` is called through `cargo run`. In other cases the path might not contain the compilation target. In these cases we take no action and just move on.

**Example**
```console
$ cargo rb hello -q
(HOST) WARN  Compilation target (thumbv8m.base-none-eabi) and core type (Armv7em) do not match. Your compilation target should be 'thumbv7em-none-eabi' (no FPU) or 'thumbv7em-none-eabihf' (with FPU).
...
```

Fixes #115.

**Todo**
- [x] Handle `CoreType::Armv8m`

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -11,10 +11,11 @@ use crate::cortexm;
 pub(crate) struct Elf<'file> {
     elf: ObjectFile<'file>,
     symbols: Symbols,
-    pub(crate) live_functions: HashSet<&'file str>,
-    pub(crate) defmt_table: Option<Table>,
-    pub(crate) defmt_locations: Option<Locations>,
+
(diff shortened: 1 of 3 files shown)


jonas-schievink:
Nice, looks great!

bors r+

PR #327, opened 2022-06-28
Optimize stack usage measuring

### LINKED ISSUE #258: Stack painting with `--measure-stack` is slow

Urhengulas:
This PR optimizes the stack usage measuring by not using the probe but a subroutine to search through the memory.

Fixes #258.

## Measurements

The speedup is similarly impressive as in #302:

| canary size | main (bcaf997cedc672ab98dc8feede19a1ca4326be05) | this PR (3ad3f86b37b56192fa6b45613ee9b9ec7b837082) |
| :---: | :---: | :---: |
| 1024 B     | 0.007s | 0.014s |
| 261060 B | 1.912s | 0.028s |

It makes sense that the time actually gets worse for a small canary size of 1024 bytes, since we are doing a lot of setup work (flash subroutine, set registers, set and reset program counter etc.). But we see that this totally pays off, since for a rather big canary of 256KiB we are almost 70 times faster!

## Further work
This PR enables us to drastically simplify the canary logic, because since both painting and measuring are pretty fast now, we can always paint the full stack.

### DIFF
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -340,7 +340,7 @@ mod measure_subroutine {
     fn search_with_probe(core: &mut Core, low_addr: u32) -> Result<Option<u32>, probe_rs::Error> {
         let mut buf = [0; self::SUBROUTINE.len()];
         core.read_8(low_addr as u64, &mut buf)?;
-        match buf.iter().position(|b| *b != CANARY_U8) {
+        match buf.into_iter().position(|b| b != CANARY_U8) {
             Some(pos) => Ok(Some(low_addr + pos as u32)),
             None => Ok(None),
         }
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -398,7 +398,7 @@ mod measure_subroutine {
     ];
 }
 
-/// Execute subroutine.
+/// Prepare and run subroutine. Also clean up afterwards.
 ///
 /// # How?
 ///
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -4,7 +4,9 @@ use probe_rs::{Core, MemoryInterface, RegisterId, Session};
 
 use crate::{registers::PC, Elf, TargetInfo, TIMEOUT};
 
+/// Canary value
 const CANARY_U8: u8 = 0xAA;
+/// Canary value
 const CANARY_U32: u32 = u32::from_le_bytes([CANARY_U8, CANARY_U8, CANARY_U8, CANARY_U8]);
 
 /// (Location of) the stack canary
@@ -194,7 +196,7 @@ macro_rules! assert_subroutine {
     };
 }
 
-/// Write [`CANARY_U32`] to the stack.
+/// Paint-stack subroutine.
 ///
 /// ### Corresponds to following rust code
 ///
@@ -230,7 +232,7 @@ macro_rules! assert_subroutine {
 mod paint_subroutine {
     use super::*;
 
-    /// Execute the subroutine.
+    /// Write [`CANARY_U32`] to the stack.
     ///
     /// ## Assumptions
     /// - Expects the [`Core`] to be halted and will leave it halted when the function
@@ -263,7 +265,7 @@ mod paint_subroutine {
     ];
 }
 
-/// Search for lowest touched address in memory.
+/// Measure-stack subroutine.
 ///
 /// ### Corresponds to following rust code
 ///
@@ -315,11 +317,10 @@ mod paint_subroutine {
 mod measure_subroutine {
     use super::*;
 
-    /// Execute the subroutine.
+    /// Search for lowest touched address in memory.
     ///
-    /// The returned `Option<u32>` is None, if the memory is untouched. Otherwise it
-    /// gives the position of the first (lowest) 4-byte-word which isn't the same as
-    /// the pattern anymore.
+    /// The returned `Option<u32>` is `None`, if the memory is untouched. Otherwise it
+    /// gives the position of the lowest byte which isn't equal to the pattern anymore.
     ///
     /// ## Assumptions
     /// - Expects the [`Core`] to be halted and will leave it halted when the function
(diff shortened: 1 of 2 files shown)


src/canary.rs
```diff
@@ -233,56 +232,165 @@ macro_rules! assert_subroutine {
...
+///
+/// ```armasm
+/// 000200ec <measure>:
+///     200ec:    4288    cmp      r0, r1
+///     200ee:    d204    bcs.n    #0xc <measure+0xe>
+///     200f0:    6803    ldr      r3, [r0, #0]
+///     200f2:    4293    cmp      r3, r2
+///     200f4:    d102    bne.n    #8 <measure+0x10>
+///     200f6:    1d00    adds     r0, r0, #4
+///     200f8:    e7f8    b.n      #0xfffffff8 <measure>
```
jonathanpallant, line 296: `#0xfffffff8` is `#-8`

src/canary.rs
```diff
@@ -233,56 +232,165 @@ macro_rules! assert_subroutine {
...
+        core.write_8(low_addr as u64, &[CANARY_U8; self::SUBROUTINE.len()])
     }
 
     const SUBROUTINE: [u8; 12] = [
         0x88, 0x42, // cmp      r0, r1
-        0x01, 0xd8, // bhi.n    200f4 <paint+0x8>
+        0x01, 0xd8, // bhi.n    #6 <paint+0x8>
         0x04, 0xc0, // stmia    r0!, {r2}
-        0xfb, 0xe7, // b.n      200ec <paint>
+        0xfb, 0xe7, // b.n      #0xfffffffa <paint>
```
jonathanpallant, line 256: `#0xfffffffa` is `#-6`

src/canary.rs
```diff
@@ -233,56 +232,165 @@ macro_rules! assert_subroutine {
...
+    }
+
+    const SUBROUTINE: [u8; 20] = [
+        0x88, 0x42, // cmp      r0, r1
+        0x04, 0xd2, // bcs.n    #0xc <measure+0xe>
+        0x03, 0x68, // ldr      r3, [r0, #0]
+        0x93, 0x42, // cmp      r3, r2
+        0x02, 0xd1, // bne.n    #8 <measure+0x10>
+        0x00, 0x1d, // adds     r0, r0, #4
+        0xf8, 0xe7, // b.n      #0xfffffff8 <measure>
```
jonathanpallant, line 374: `#-8` is clearer I think

src/canary.rs
```diff
@@ -217,9 +216,9 @@ macro_rules! assert_subroutine {
 /// ```armasm
 /// 000200ec <paint>:
 ///    200ec:    4288    cmp      r0, r1
-///    200ee:    d801    bhi.n    200f4 <paint+0x8>
+///    200ee:    d801    bhi.n    #6 <paint+0x8>
 ///    200f0:    c004    stmia    r0!, {r2}
-///    200f2:    e7fb    b.n      200ec <paint>
+///    200f2:    e7fb    b.n      #0xfffffffa <paint>
```
jonathanpallant, line 221: `#-6` is clearer here I think

src/canary.rs
```diff
@@ -217,9 +216,9 @@ macro_rules! assert_subroutine {
 /// ```armasm
```
jonathanpallant, line 216: The comment above should be changed to match the one below.

src/canary.rs
```diff
@@ -233,56 +232,165 @@ macro_rules! assert_subroutine {
...
+        match self::search_with_probe(core, low_addr)? {
+            addr @ Some(_) => return Ok(addr), // if we find a touched value, return early ...
+            None => {}                         // ... otherwise we continue
+        }
+
+        super::execute_subroutine(core, low_addr, stack_size, self::SUBROUTINE)?;
+        self::get_result(core)
+    }
+
+    fn search_with_probe(core: &mut Core, low_addr: u32) -> Result<Option<u32>, probe_rs::Error> {
```
jonathanpallant, line 340: I'd like to see comments for these private functions, for the benefit of the next maintainer.

```
/// Searches though memory byte by byte using the SWD/JTAG probe.
///
/// We need this because we're about to overwrite some stack with a
/// canary-measuring program, but we need to check the memory we're about
/// to overwrite in case the stack usage got this far.
```

src/canary.rs
Urhengulas: Nice. I was thinking about that, but wasn't sure if that is "allowed".

src/canary.rs
```diff
@@ -233,56 +232,165 @@ macro_rules! assert_subroutine {
...
+///         }
+///     }
+///
+///     result
+/// }
+/// ```
+///
+/// ### Generated assembly
+///
+/// The assembly is generated from aboves rust code, using the jorge-hack.
```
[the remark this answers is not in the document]
Urhengulas: Not yet, but I plan to write a blog post about it.

Essentially we create a new cargo project, write our function in there, set the cargo profile to optimize for size, compile that for `thumbv6m-none-eabi` and then inspect the final binary with `objdump`. We do that so we don't have to play optimizer.

src/canary.rs
jonathanpallant: I think you can do what you like in the comments :)

Urhengulas: That's true. I just thought I would like to keep it in a format an assembler would accept. So people don't need to guess what I meant, but can just read it up in some documentation.

I just tried, but I couldn't figure out any way to get this to work with `arm-none-eabi-as`, neither with `#0xfffffff8`, nor with `#-8`. So I will just go with `#-8`, because, as you said, it is more clear.

src/canary.rs
Urhengulas: I've documented all functions. What do you think?

src/canary.rs
```diff
@@ -210,79 +212,219 @@ macro_rules! assert_subroutine {
 /// }
 /// ```  
 ///
-/// ### Generated assembly
+/// # Assembly
 ///
-/// The assembly is generated from aboves rust code, using the jorge-hack.
+/// The assembly is generated from the Rust function `fn measure()` above, using the
```
Urhengulas, line 217:
```suggestion
/// The assembly is generated from the Rust function `fn pain()` above, using the
```

jonathanpallant (approved):

jonathanpallant:
Great work. Ship it!!

PR #330, opened 2022-07-14
Fix `fn round_up`.

Urhengulas:
It was hardcoding `k` as `4`. This PR changes the fn to actually use `k`.
This resultet in no problem so far, because it was only used with `k=4`.

### DIFF
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -169,6 +169,7 @@ impl Canary {
     }
 }
 
+/// Rounds up to the next multiple of `k` that is greater or equal to `n`.
 fn round_up(n: u32, k: u32) -> u32 {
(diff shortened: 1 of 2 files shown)


src/canary.rs
```diff
@@ -174,7 +174,7 @@ fn round_up(n: u32, k: u32) -> u32 {
     if rem == 0 {
         n
     } else {
-        n + 4 - rem
+        n + k - rem
```
justahero, line 177: Even though it's a private function, could you document it? Imho it's not immediately obvious that the function rounds up to next multiple of the given value `k`.

Urhengulas:
bors r=justahero

PR #331, opened 2022-07-20
Refactor stack painting

Urhengulas:
Most notably:
- set registers using probe-rs, not in the subroutine anymore
  → this enables to have the subroutine as a const array
- bundle assertions in `assert_subroutine!` macro
  → macro and not fn to preserve correct location in the panic message
- encapsulate function and subroutine in a module
- greatly improve documentation
- rename arguments `start` and `end` to `low_addr` and `high_addr`
- use `low_addr` and `stack_size` as arguments

(This PR can be considered a preparation for #327)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - [#293] Update snapshot tests to new TRACE output
 
 [#331]: https://github.com/knurling-rs/probe-run/pull/331
+[#330]: https://github.com/knurling-rs/probe-run/pull/330
 [#329]: https://github.com/knurling-rs/probe-run/pull/329
 [#328]: https://github.com/knurling-rs/probe-run/pull/328
 [#326]: https://github.com/knurling-rs/probe-run/pull/326
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+- [#xxx] Refactor stack painting
 - [#330] Fix `fn round_up`
 - [#329] Update probe-rs to 0.13.0 (does not yet implement 64-bit support)
 - [#328] Simplify, by capturing identifiers in logging macros
@@ -17,7 +18,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - [#314] Clarify documentation in README
 - [#293] Update snapshot tests to new TRACE output
 
-[#330]: https://github.com/knurling-rs/probe-run/pull/330
+[#xxx]: https://github.com/knurling-rs/probe-run/pull/xxx
 [#329]: https://github.com/knurling-rs/probe-run/pull/329
 [#328]: https://github.com/knurling-rs/probe-run/pull/328
 [#326]: https://github.com/knurling-rs/probe-run/pull/326
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
-- [#xxx] Refactor stack painting
+- [#331] Refactor stack painting
 - [#330] Fix `fn round_up`
 - [#329] Update probe-rs to 0.13.0 (does not yet implement 64-bit support)
 - [#328] Simplify, by capturing identifiers in logging macros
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+- [#xxx] Refactor stack painting
```
Urhengulas, line 9:
```suggestion
- [#331] Refactor stack painting
```

CHANGELOG.md
```diff
@@ -17,7 +18,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - [#314] Clarify documentation in README
 - [#293] Update snapshot tests to new TRACE output
 
-[#330]: https://github.com/knurling-rs/probe-run/pull/330
+[#xxx]: https://github.com/knurling-rs/probe-run/pull/xxx
```
Urhengulas, line 21:
```suggestion
[#331]: https://github.com/knurling-rs/probe-run/pull/331
```

CHANGELOG.md
```diff
@@ -17,7 +18,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - [#314] Clarify documentation in README
 - [#293] Update snapshot tests to new TRACE output
 
-[#330]: https://github.com/knurling-rs/probe-run/pull/330
```
jonathanpallant, line 20: Did you mean to remove #330?

src/canary.rs
```diff
@@ -179,88 +180,124 @@ fn round_up(n: u32, k: u32) -> u32 {
...
+///     while low_addr <= high_addr {
+///         (low_addr as *mut u32).write(pattern);
+///         low_addr += 4;
+///     }
+/// }
+/// ```  
+///
+/// ### Generated assembly
+///
+/// The assembly is generated from abothes rust code, using the jorge-hack. The label
```
jonathanpallant, line 215: abothes?

src/canary.rs
```diff
@@ -179,88 +180,124 @@ fn round_up(n: u32, k: u32) -> u32 {
...
-        0x04, 0x4a, // ldr r2, [pc, #16]
-        // <loop>
-        0x81, 0x42, // cmp r1, r0
-        0x01, 0xD0, // beq.n   116 <end>
-        0x04, 0xC0, // stmia   r0!, {r2}
-        0xFB, 0xE7, // b.n 10e <loop>
+    const SUBROUTINE: [u8; 12] = [
+        // <paint>
+        0x88, 0x42, // cmp      r0, r1
+        0x01, 0xd8, // bhi.n    200f4 <end>
```
jonathanpallant, line 264: If this is a relative jump, the comment should probably indicate that (the comment looks like an absolute address?)

jonathanpallant (changes requested):
Some queries to answer, but otherwise looks good.

CHANGELOG.md
Urhengulas: Nope, I didn't 😅

src/canary.rs
Urhengulas: Oh, should be "aboves"

src/canary.rs
Urhengulas: The actual comment was `bhi.n	200f4 <paint+0x8>`, which indicates the relative jump. I though that it is slightly easier to read when I change it, but apparently it creates more confusion.

Urhengulas:
bors r=jonathanpallant

src/canary.rs
jonathanpallant: It's the 200f4 that confuses me, because that seems like an absolute value.

Urhengulas: Okay. I will investigate

Urhengulas: I assume that the problem with the address being absolute would be, that we jump to the wrong location in memory and therefore our subroutine would crash. Is this correct?

I've played around with it and placed the subroutine in different memory locations. It always worked as expected. Is there another way I can test if this is a problem?

Urhengulas: Oh no. You are right :(

jonathanpallant: I think the instruction is a relative jump, but the disassembly has translated it into an absolute address based on the location? I don't know thumbv7m well enough to disassemble that instruction in my head, but a 16 bit instruction seems too short for an absolute address.

Urhengulas: Puh. The failure I observed earlier was due to another problem I introduced while testing.

This [online disassembler](http://shell-storm.org/online/Online-Assembler-and-Disassembler/?opcodes=02d1&arch=arm-t&endianness=little&dis_with_raw=True&dis_with_ins=True#disassembly) decodes `01d8` as `bhi #6`, which looks like a relative jump to me. So it really seems that objdump resolves that address. I couldn't find how to instruct it to not do that.

Urhengulas: I'd err for keeping the comment like it is, since it is what objdump outputs. We maybe could add a note that it is a relative jump after all.

jonathanpallant: But objdump only says that because you told objdump where the data was located in memory when you disassembled it. This is documentation for piece of relocatable code that can be placed anywhere in memory (it goes in at the bottom of the stack region being painted, right?). So personally I would change it to be `bhi #6 <paint+0x08>`.

Urhengulas: Also true. I will include that in #327.

PR #344, opened 2022-08-10
Replace `pub(crate)` with `pub`

Urhengulas:
Since `probe-run` is a binary crate, `pub` has the same effect as `pub(crate)`.


Urhengulas:
bors retry

PR #357, opened 2022-10-06
Update to `clap 4.0`

Urhengulas:
This PR updates our cli parser `clap` from `2.34` to `4.0`.

The end-user experience should stay similar, but has a fresh UI, many internal improvements, and better help messages when a user gets the arguments wrong.

On the left you can see the new UI, and on the right the old.
image

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -24,7 +24,7 @@ pub struct Opts {
     pub backtrace_limit: u32,
 
     /// The chip to program.
-    #[arg(long, required = true, conflicts_with_all = HELPER, env = "PROBE_RUN_CHIP")]
+    #[arg(long, required = true, conflicts_with_all = HELPER_CMDS, env = "PROBE_RUN_CHIP")]
     chip: Option<String>,
 
     /// Path to chip description file, in YAML format.
@@ -40,7 +40,7 @@ pub struct Opts {
     pub disable_double_buffering: bool,
 
     /// Path to an ELF firmware file.
-    #[arg(required = true, conflicts_with_all = HELPER)]
+    #[arg(required = true, conflicts_with_all = HELPER_CMDS)]
     elf: Option<PathBuf>,
 
     /// Output logs a structured json.
@@ -97,7 +97,7 @@ pub struct Opts {
 }
 
 /// Helper commands, which will not execute probe-run normally.
-const HELPER: [&str; 3] = ["list_chips", "list_probes", "version"];
+const HELPER_CMDS: [&str; 3] = ["list_chips", "list_probes", "version"];
 
 pub fn handle_arguments() -> anyhow::Result<i32> {
     let opts = Opts::parse();
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -24,7 +24,7 @@ pub struct Opts {
     list_probes: bool,
 
     /// The chip to program.
-    #[arg(long, required_unless_present_any(&["list_chips", "list_probes", "version"]), env = "PROBE_RUN_CHIP")]
+    #[arg(long, required_unless_present_any(HELPER), env = "PROBE_RUN_CHIP")]
     chip: Option<String>,
 
     /// Path to chip description file, in YAML format.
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -1,102 +1,105 @@
...
 
     /// Enable more verbose output.
-    #[structopt(short, long, parse(from_occurrences))]
-    pub verbose: u32,
+    #[arg(short, long, action = ArgAction::Count)]
+    pub verbose: u8,
 
     /// Prints version information
-    #[structopt(short = "V", long)]
+    #[arg(short = 'V', long)]
```
Dajamante, line 63: Did not notice before that 'V' != 'v' !

src/cli.rs
```diff
@@ -1,102 +1,105 @@
...
+    #[arg(long)]
     pub verify: bool,
 
     /// Arguments passed after the ELF file path are discarded
-    #[structopt(name = "REST")]
+    #[arg(allow_hyphen_values = true, hide = true, trailing_var_arg = true)]
     _rest: Vec<String>,
 }
 
+/// Helper commands, which will not execute probe-run normally.
```
Dajamante, line 98: So you can pass `probe-run HELPER`?
I don't get it from the documentation.

Urhengulas:
> Also it is not in alphabetic order anymore, so one need a few moments of looking up and down...

I was thinking about that as well. The reasoning of the authors of clap is that they want to give the cli authors (us) control over the order. We can just solve this by alphabetically ordering the args.

src/cli.rs
```diff
@@ -1,102 +1,105 @@
...
     list_chips: bool,
 
     /// Lists all the connected probes and exit.
-    #[structopt(long)]
+    #[arg(long)]
     list_probes: bool,
 
     /// The chip to program.
-    #[structopt(long, required_unless_one(&["list-chips", "list-probes", "version"]), env = "PROBE_RUN_CHIP")]
+    #[arg(long, required = true, conflicts_with_all = HELPER, env = "PROBE_RUN_CHIP")]
```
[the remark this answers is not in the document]
Urhengulas: Not at all. It conflicts with `--list-chips`, `--list_probes`, `--version`. It seems I need a better name for `HELPER`.

src/cli.rs
Urhengulas: Yes. `-v` is `--verbose` and `-V` is `--version`.

src/cli.rs
```diff
@@ -1,102 +1,105 @@
...
-    #[structopt(name = "ELF", parse(from_os_str), required_unless_one(&["list-chips", "list-probes", "version"]))]
+    #[arg(required = true, conflicts_with_all = HELPER)]
     elf: Option<PathBuf>,
 
     /// Skip writing the application binary to flash.
-    #[structopt(
+    #[arg(
         long,
-        conflicts_with = "disable-double-buffering",
+        conflicts_with = "disable_double_buffering",
```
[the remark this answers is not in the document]
Urhengulas: This flag changes the flashing behavior of probe-rs: https://docs.rs/probe-rs/latest/probe_rs/flashing/struct.DownloadOptions.html#structfield.disable_double_buffering

src/cli.rs
Urhengulas: No. `HELPER` is just a helper variable to not have `["list_chips", "list_probes", "version"]` twice. What could be a better name?

src/cli.rs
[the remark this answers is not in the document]
Urhengulas: What do you think about `HELPER_CMDS`?

Urhengulas:
bors r=Dajamante

src/cli.rs
Dajamante: I think `HELPER_CMD` is good! I think I got confused because I expected to see `<HELPER_CMD>`, to recognize it immediately as an argument. But that's just me.

Dajamante (commented):
Thank you for addressing my questions.
Another quick one: why do we need `Cargo.lock`, and can we put `<HELPER_CMD>` between arrows, or is it against conventions?

src/cli.rs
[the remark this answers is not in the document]
Dajamante: Ah ok I see it now when comparing again with `structopt`. Those three conflict : `list-chips`, `list-probes`, `version`.

Urhengulas:
> why do we need `Cargo.lock`?

Afaik it is convention to commit the lock file for applications, but not for libraries.

> can we put `<HELPER_CMD>` between arrows, or is it against conventions?

That would be possible.

PR #373, opened 2023-01-17
Update to probe-rs 0.14

korken89:
Update to latest probe-rs


Urhengulas:
Thank you @korken89!

bors r+

PR #380, opened 2023-02-11
`--erase-all` to erase all nonvolatile memory

### LINKED ISSUE #368: Erase-all flash option
Would an option to erase all flash be considered generally useful? The gen 3 nRF52 boards have AP Protect enabled by default, so it might be handy for an optional  erase all to enable subsequent flashing.

Urhengulas:
Fixes #368

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -62,7 +62,10 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let probe = probe::open(opts)?;
 
     let probe_target = target_info.probe_target.clone();
-    let permissions = Permissions::default().allow_erase_all();
+    let permissions = match opts.erase_all {
+        false => Permissions::new(),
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -62,10 +62,11 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let probe = probe::open(opts)?;
 
     let probe_target = target_info.probe_target.clone();
+    let permissions = Permissions::default().allow_erase_all();
```
jonathanpallant, line 65: Could you avoid obtaining this permission if you are not going to erase all?

jonathanpallant (changes requested):
Left a comment about permissions.

src/main.rs
Urhengulas: Yes, makes sense!

PR #383, opened 2023-02-17
Fix unwinding

### LINKED ISSUE #382: Consider removing the need for a "sentinel stack frame" while unwinding
At the moment, it is necessary for `probe-run` to "run into" a sentinel frame where the link register is recorded as `0xFFFF_FFFF` while unwinding (on panic).

https://github.com/rust-embedded/cortex-m-rt/pull/337 was opened to include this in `c-m-rt`, but it turns out, this caused [a pretty serious regression](https://gist.github.com/jamesmunns/7ee8d1891cb44b5f81d9c11d4569b2fb), that we just recently caught.

We're fixing this for `cortex-m-rt` v0.7.3 (in https://github.com/rust-embedded/cortex-m/pull/463) by pushing another word to the stack, but this means that all targets have 8 bytes of "dead" stack usage, as well as some startup code (and confusing asm) necessary to insert the necessary "full stack frame".

During testing, we noticed that GDB was actually totally happy to not have the sentinel frame at all - only `probe-run` was upset (it throws a warning that the stack is corrupted).

We're currently looking at whether it would be possible to totally remove the sentinel frame in c-m-rt. Ideally, this would involve probe-run being able to handle the case where it hits the end of the stack without hitting the sentinel frame.

No hard decision has been made to remove the termination frame (and cfi/cfa hints) yet, but there seems to be some desire to. We decided to keep it (and add a second push) for now, to allow us to ship 0.7.3 ASAP, and not break (or cause problems for) probe-run.

---

(suggested impl ahead, feel free to ignore): I *think* it should be possible to notice when hitting the stack top symbol, and if it was hit "gracefully", e.g. there is SOME frame right at the top of the stack with no excess data, it can be treated as the terminal frame, and not throw a "stack corrupted" warning.

I'm happy to provide a patched version of cortex-m-rt that would reproduce the currently discussed behavior for testing.

Urhengulas:
This PR "fixes" unwinding, by not relying on the sentinel stack frame anymore, but instead checking if unwinding ends in the `Reset` function. If unwinding ends in the `Reset` function, it is considered NOT corrupted; if it ends somewhere else, it is considered corrupted.

Fixes #382 

Todo:
- ❌ ~~Is the `Reset` symbol always present?~~ -> No, "an unmangled symbol named `Reset` is even less common than main"
- ❌  ~~Does any thumb bit need to be cleared?~~ -> no
- ❌  ~~Test it with an corrupted stack~~ -> not trivially possible from Rust

### DIFF
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -95,6 +95,10 @@ pub fn target(
 
         log::debug!("LR={lr:#010X} PC={pc:#010X}");
 
+        // Link Register contains an EXC_RETURN value. This deliberately also includes
+        // invalid combinations of final bits 0-4 to prevent futile backtrace re-generation attempts
+        let exception_entry = lr >= cortexm::EXC_RETURN_MARKER;
+
         let program_counter_changed = !cortexm::subroutine_eq(lr, pc);
 
         // If the frame didn't move, and the program counter didn't change, bail out
@@ -104,10 +108,6 @@ pub fn target(
             output.corrupted = !reset_range.contains(&pc);
             break;
         }
-
-        // Link Register contains an EXC_RETURN value. This deliberately also includes
-        // invalid combinations of final bits 0-4 to prevent futile backtrace re-generation attempts
-        let exception_entry = lr >= cortexm::EXC_RETURN_MARKER;
 
         if exception_entry {
             output.raw_frames.push(RawFrame::Exception);
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -93,12 +93,16 @@ pub fn target(core: &mut Core, elf: &Elf, active_ram_region: &Option<RamRegion>)
 
         let program_counter_changed = !cortexm::subroutine_eq(lr, pc);
 
-        // If the frame didn't move, and the program counter didn't change, bail out (otherwise we
-        // might print the same frame over and over).
-        output.corrupted = !cfa_changed && !program_counter_changed;
+        match !cfa_changed && !program_counter_changed {
+            // If the frame didn't move, and the program counter didn't change, bail out
+            // (otherwise we might print the same frame over and over).
+            true => {
+                // If we do not end up in the reset function the stack is corrupted
+                output.corrupted = !elf.reset_fn_range().contains(&pc);
 
-        if output.corrupted {
-            break;
+                break;
+            }
+            false => output.corrupted = false,
         }
 
         if exception_entry {
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -147,10 +147,12 @@ pub fn target(
         }
 
         // if the SP is above the start of the stack, the stack is corrupt
-        let advanced_sp = registers.get(registers::SP).unwrap();
-        if advanced_sp > stack_start {
-            output.corrupted = true;
-            break;
+        match registers.get(registers::SP) {
+            Ok(advanced_sp) if advanced_sp > stack_start => {
+                output.corrupted = true;
+                break;
+            }
+            _ => {}
         }
     }
 
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -1,5 +1,7 @@
 //! unwind target's program
 
+use std::cmp::Ordering;
+
 use anyhow::{anyhow, Context as _};
 use gimli::{
     BaseAddresses, CieOrFde, DebugFrame, FrameDescriptionEntry, Reader, UnwindContext,
@@ -39,9 +41,6 @@ pub fn target(
         raw_frames: vec![],
         processing_error: None,
     };
-
-    dbg!(stack_start, active_ram_region);
-    println!("{:#x}", stack_start);
 
     /// Returns all info collected until the error occurred and puts the error into `processing_error`
     macro_rules! unwrap_or_return_output {
@@ -133,20 +132,24 @@ pub fn target(
             output.processing_error = Some(anyhow!(
                 "bug? LR ({lr:#010x}) didn't have the Thumb bit set",
             ));
-            return output;
+            break;
         }
 
         // check if we want to end unwinding
         let advanced_sp = registers.get(registers::SP).unwrap();
-
-        // if the stack pointer is at the start of the stack, exit gracefully
-        if advanced_sp == stack_start {
-            output.corrupted = false;
-            return output;
-        // if the stack pointer is at the start of the stack, the stack is corrupt
-        } else if advanced_sp > stack_start {
-            output.corrupted = true;
-            return output;
+        match advanced_sp.cmp(&stack_start) {
+            // if the SP is at the start of the stack, unwinding is completed
+            Ordering::Equal => {
+                output.corrupted = false;
+                break;
+            }
+            // if the SP is above the start of the stack, the stack is corrupt
+            Ordering::Greater => {
+                output.corrupted = true;
+                break;
+            }
+            // if the SP is below the start of the stack, continue unwinding
+            Ordering::Less => continue,
         }
     }
 
--- a/src/backtrace/unwind.rs
+++ b/src/backtrace/unwind.rs
@@ -1,5 +1,7 @@
 //! unwind target's program
 
+use std::ops::Range;
+
 use anyhow::{anyhow, Context as _};
 use gimli::{
     BaseAddresses, CieOrFde, DebugFrame, FrameDescriptionEntry, Reader, UnwindContext,
@@ -32,6 +34,7 @@ pub fn target(
     elf: &Elf,
     active_ram_region: &Option<RamRegion>,
     stack_start: u32,
+    reset_range: Range<u32>,
 ) -> Output {
     let mut output = Output {
         corrupted: true,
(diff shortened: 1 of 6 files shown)


japaric:
an unmangled symbol named `Reset` is even less common than `main` (which at least is common in C programs). neither symbol is specified in the ABI of the [cortex-m-rt crate](https://docs.rs/cortex-m-rt/0.7.3/cortex_m_rt/index.html) so relying on either could lead to breakage in the future ("future" being any new patch version, i.e. 0.7.x).

on the other hand, `_stack_start` is part of the ABI of the cortex-m-rt crate so [this idea](https://github.com/knurling-rs/probe-run/issues/382#issuecomment-1433131110) may work at least for cortex-m-rt v0.7.x but the ABI could change in 0.8.0.

another way to figure out the address of the reset handler is to read the second entry in the vector table. [this document](https://developer.arm.com/documentation/dui0552/a/the-cortex-m3-processor/exception-model/vector-table) states:

> On system reset, the vector table is fixed at address `0x00000000`.

thought I'm not sure that is specified in the Cortex-M ISA. I know STM32 devices use `0x2000_0000` as the start address of Flash memory so firmware put the vector table at that address, instead of at `0x0000_0000`. It could be that the Flash memory at `0x2000_0000` is aliased at address `0x0000_0000` but I have not checked.

yet another way to get the address of the reset handler is to issue a reset-halt and read out the program counter (PC) register. I don't have a link to normative documentation but the Cortex-M ISA states that the program counter is set to the second entry of the vector table (the address of the reset handler) after a system reset. I think this runtime approach would be the sure way to get the address. after a reset-halt you could read the SP register and that would give you `_stack_start` as well.

Urhengulas:
Thank you for the answer @japaric. A few clarification questions:
1. Are you suggesting doing both the `_stack_start` idea James and checking if we end up in the reset handler? Or are you suggesting that we do one or the other?
2. Obtaining the reset handler during runtime sounds good. But as far as I understand it only gives me the address of the reset handler, not the size. In order to check if we end up inside of it I need both (as far as I understand). How can I get the size?

🥂

Urhengulas:
I've switched to James proposal of ending the unwinding depending on the `_stack_start` and that seems to work as well (see https://github.com/knurling-rs/probe-run/pull/383/commits/3e7fe2d7d693baff4f6597b4a245a164c6566289).

Atm I extract the stack start by reading the SP after a reset (as you've proposed). It seems to me that `_stack_start` is the same as the start of the `active_ram_region` we extract from the ELF. Is this just a coincidence, or always the case? If they are same we can just do one and not both.

japaric:
> Are you suggesting doing both the _stack_start idea James and checking if we end up in the reset handler? Or are you suggesting that we do one or the other?

I think it'd be good to check the stack pointer (SP) value against `_stack_start` during each step of the unwinding process. if the SP value ever goes higher than `_stack_start` then either the stack is corrupted or there's a bug in the unwinding logic; in either case we should stop the unwinding process.

Thinking more about it, I think it may be better not to use `_stack_start == SP` as the stop condition. reason being, the reset handler could adjust the stack pointer (e.g. to make it multiple of 4 or 8 bytes so that the *next* stack frame adheres to e.g. the AAPCS ABI) before jumping into the next function. the reset handler itself usually does not follow any ABI (as it's written in assembly) and could lack both a function prologue and enough CFI (call frame information) to let the unwinder compute the value of SP at the *start* of the reset handler so it may not be possible (for the unwinder) to compute that SP was equal to `_stack_start` at some point in time.

> But as far as I understand it only gives me the address of the reset handler, not the size. In order to check if we end up inside of it I need both (as far as I understand)

correct.

> . How can I get the size?

you should be able to iterate the symbol table to find the size (the API may be called `symbols` or `symbol_table` depending on whether you are using the `xmas_elf` or `object` library). each symbol entry has an associated start address and size (+) so you can compute the address range each symbol spans (`start .. (start+size)`). you can then use `reset_handler_address_span.contains(current_pc)` as the stop condition.

(+) should look bit like `nm -CSn` if you print the data

``` console
$ # first column is start address; second is size; numbers are in hexadecimal format
$ arm-none-eabi-nm -CSn some-elf
(..)
00000100 00000028 T Reset
(..)
0000013c 0000000a T main
00000148 00000028 t hello::__cortex_m_rt_main
00000170 0000000a T app::exit
```

---

you asked earlier:

>  Does any thumb bit need to be cleared?

you can pretty much always clear the thumb bit (bit 0) when comparing addresses. the start address of a function (as seen in the `objdump` output) must always be an even number; the value of PC will always be an even number. 

values that eventually end up in an instructions like `bl r0` ("branch into function whose address is stored in register R0") have to have the thumb bit set. as those values (effectively function pointer values) are stored in static variables, they are usually stored with the bit already set to avoid setting it at runtime. when you read the second entry of the vector table (either at runtime or from the rodata stored in the ELF) you'll get an odd number.

Urhengulas:
@japaric said:
> Thinking more about it, I think it may be better not to use `_stack_start == SP` as the stop condition.

What then? That unwinding ends in the reset handler? This is what I implemented in 9f40539e2d537c49ea22eb1ef7cae5f6d50b4fd6.

If this is what you mean we would have:
- a successful stop if unwinding ends (`!cfa_changed && !program_counter_changed`) in the reset handler
- unsuccessful stop if
  - unwinding stops outside of reset handler
  - `SP > _stack_start`

Correct?

---

@japaric said:
> @Urhengulas said:
> > How can I get the size?
> 
> you should be able to iterate the symbol table to find the size

Didn't you say earlier that we cannot rely on the symbol table? I talking about following quote.

@japaric said:
> an unmangled symbol named `Reset` is even less common than `main` (which at least is common in C programs). neither symbol is specified in the ABI of the [cortex-m-rt crate](https://docs.rs/cortex-m-rt/0.7.3/cortex_m_rt/index.html)

Or am I misunderstanding something? Can we figure out the mangled name of the symbol?

Urhengulas:
@japaric The PR is ready for review again. I think it makes the code oven messier than it already is, and I am working on a follow-up PR to clean things up. But I wouldn't do this in here.

src/main.rs
```diff
@@ -456,3 +468,22 @@ fn flashing_progress() -> flashing::FlashProgress {
         }
     })
 }
+
+fn get_stack_start_and_reset_handler(
+    sess: &mut Session,
+    elf: &Elf,
+) -> anyhow::Result<(u32, Range<u32>)> {
+    let mut core = sess.core(0)?;
+    core.reset_and_halt(Duration::from_secs(5))?;
```
japaric, line 477: is it necessary to do a reset halt *here*? there's another call to reset_and_halt in `main.rs` due to `Canary::install` so this may result in the device being reset twice, which should not be a problem to the operation of probe-run but it's preferable to not do double work.

perhaps the reset_and_halt can be do once before calling this function, then it can also be removed from canary::install

src/backtrace/unwind.rs
```diff
@@ -87,21 +95,17 @@ pub fn target(core: &mut Core, elf: &Elf, active_ram_region: &Option<RamRegion>)
...
-        // If the frame didn't move, and the program counter didn't change, bail out (otherwise we
-        // might print the same frame over and over).
-        output.corrupted = !cfa_changed && !program_counter_changed;
-
-        if output.corrupted {
+        // If the frame didn't move, and the program counter didn't change, bail out
+        // (otherwise we might print the same frame over and over).
+        if !cfa_changed && !program_counter_changed {
+            // If we do not end up in the reset function the stack is corrupted
+            output.corrupted = !reset_range.contains(&pc);
```
japaric, line 108: this is not due to this PR but I wouldn't necessarily categorize this exit condition as "corrupted". if you compile the rust program with `-C force-frame-pointers=off` then it's not possible to unwind the stack because frame pointer information is missing (from register r7) so you'll hit this branch but the call stack and the program will both operate fine.

`-C force-frame-pointers=off` is not very common but the assembly trampolines used to implement context switches in RTOSes / kernels could have the same effect of making unwinding not work.

I think it would be more to correct to simply say in this branch that it's "not possible to further unwind the stack" or something like that

src/backtrace/unwind.rs
```diff
@@ -139,7 +143,14 @@ pub fn target(core: &mut Core, elf: &Elf, active_ram_region: &Option<RamRegion>)
             output.processing_error = Some(anyhow!(
                 "bug? LR ({lr:#010x}) didn't have the Thumb bit set",
             ));
-            return output;
+            break;
+        }
+
+        // if the SP is above the start of the stack, the stack is corrupt
+        let advanced_sp = registers.get(registers::SP).unwrap();
```
japaric, line 150: this should be an `if let` to avoid a potential panicking branch.

src/main.rs
```diff
@@ -456,3 +468,22 @@ fn flashing_progress() -> flashing::FlashProgress {
...
+    let stack_start = core.read_core_reg::<u32>(SP)?;
+    let reset_address = cortexm::set_thumb_bit(core.read_core_reg::<u32>(PC)?);
+    core.reset()?;
+    let reset = elf
```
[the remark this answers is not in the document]
Urhengulas: In case there is not exactly one result we should abort, right?

src/main.rs
Urhengulas: I will address this in the follow-up clean-up PR, since I will change the order of operations there and also touch the canary

src/backtrace/unwind.rs
Urhengulas: But there always will be a stack pointer value because it gets added when constructing the `Registers`.

Urhengulas: I changed it anyways

src/backtrace/unwind.rs
Urhengulas: I opened https://github.com/knurling-rs/probe-run/issues/387 for this.

japaric (approved):

Urhengulas:
bors r=japaric

PR #388, opened 2023-03-08
Draft: Start a GDB server with --gdblisten 127.0.0.1:1337

jonathanpallant:
TODO: Turn down some warning logs
TODO: Stop the GDB server when probe-run exits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ log = "0.4"
 object = { version = "0.30", default-features = false }
 probe-rs = "0.17"
 signal-hook = "0.3"
+gdb-server = "0.17"
 
 [dev-dependencies]
 dirs = "4"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -91,6 +91,10 @@ pub struct Opts {
     #[arg(long)]
     pub verify: bool,
 
+    /// Starts a GDB server on this port
+    #[arg(long)]
+    pub gdblisten: Option<String>,
+
     /// Prints version information
     #[arg(short = 'V', long)]
     version: bool,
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -27,6 +27,7 @@ log = "0.4"
 object = { version = "0.30", default-features = false }
 probe-rs = "0.17"
 signal-hook = "0.3"
+gdb-server = "0.17"
```
Urhengulas, line 30: Please keep the alphabetical ordering. Makes it easier to scan through it.

src/cli.rs
```diff
@@ -91,6 +91,23 @@ pub struct Opts {
     #[arg(long)]
     pub verify: bool,
 
+    /// Sets which port we bind the GDB Server to.
+    ///
+    /// Note: doesn't necessarily start the GDB Server. See `--gdb-on-start` and
+    /// `--gdb-on-crash`.
+    #[arg(long, env = "PROBE_RUN_GDB_PORT")]
+    pub gdb_port: Option<String>,
```
Urhengulas, line 99: I wondered if this should be `Option<SocketAddr>` instead? This will give users a suitable error message if they provide an invalid address.

```console
$ cargo run -- --gdbport 127.0.0:3333
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/krate --gdbport '127.0.0:3333'`
error: invalid value '127.0.0:3333' for '--gdbport <GDBPORT>': invalid socket address syntax

For more information, try '--help'.
```

But, since `GdbInstanceConfiguration::from_session` takes a `Option<impl Into<String>>`, I am unsure.

jonathanpallant: That's a good idea. We can always turn the valid socket address back into a string before giving it to the GDB server.

jonathanpallant: Also we should warn if they provide a gdb-port but fail to activate either of the GDB options (-on-start or -on-fault)

Issue #323, opened 2022-05-23
Can't find room for painting the stack when stack isn't in RAM region


jonathanpallant:
**Describe the bug**

I fell over a bug in the stack painting stuff working on Neotron. I don't think `probe-run` can cope when the stack isn't in the RAM region. In my case I have a RAM region for .data/.bss, and two separate stacks, in regions called `CORE0_STACK` and `CORE1_STACK`. It tries to fit the painting code into RAM and gets sad when it doesn't fit, when `CORE0_STACK` is right there empty.

It's annoying because the failure mode isn't "I didn't paint your stack", it's "I exited and refused to take your chip out of reset".

**To Reproduce**

Make a project where stack isn't in the RAM region, but is in another region (like `STACK`). Set `__stack_start = ORIGIN(STACK) + LEN(STACK)` so cortex-m-rt does the right thing.

**Expected and observed behavior**

I expected `probe-run` to flash my chip and start my code.

What it actually says it that it couldn't find the space for the painting code, and exited.

**config.toml**

See https://github.com/Neotron-Compute/Neotron-Pico-BIOS/tree/12f735402400b421d4e0fb3a3bd717b34fe84a40, but you'll want to make the RAM region smaller to trigger the bug (it has a bunch of padding in it currently).

**Probe details**

It's an `RP2040` with the DapperMime firmware.

**Operating System:**

Arch Linux.

**ELF file (attachment)**

Can attach later, or you can build it from above repo.

**Additional context**

N/A

Urhengulas:
https://github.com/knurling-rs/probe-run/issues/39 is related, as it is also concerned with multiple ram regions.

PR #410, opened 2023-06-13
Simplify canary

### LINKED ISSUE #39: Support stack overflow detection on chips with multiple RAM regions
The automatic stack canary currently will be disabled when the chip has multiple memory regions. We should try to at least support the case where all but one region are unused by the app.

Urhengulas:
This PR simplifies our canary system. This is possible now, because our stack-painting and stack-measuring got blazingly fast due to #302 #327 #331.

Adds work around for #323.

# What

Before this PR there were two modes. By default `probe-run` would only paint 10% of stack and use that to detect potential stack-overflows. If a user would pass `--measure-stack`, `probe-run` would paint the whole stack (which was slow) and report the percentage of stack used, but don't notify about potential stack-overflows.

After this PR there is only a single mode. `probe-run` always paints the whole stack (which is fast), always reports the percentage of stack used and always reports potential stack-overflows.

The `--measure-stack` option is not completely removed to avoid a breaking change. But it does not do anything anymore except emitting a deprecation warning.

Some other drive-by changes:
- assert subroutine requirements more early, therefore work around #323
- refactor `Settings` to not need a lifetime

# Testing

The unit testing[^1] and on-device tests[^2] succeed

# Note to reviewer

While the whole PR got kinda big, the individual commits should be rather self-contained and therefore easier to review.

[^1]: `cargo test`
[^2]: `cargo test -- --ignored`

### DIFF
--- a/src/canary.rs
+++ b/src/canary.rs
@@ -2,7 +2,11 @@ use std::time::Instant;
 
 use probe_rs::{Core, MemoryInterface, RegisterId};
 
-use crate::{registers::PC, Elf, TargetInfo, TIMEOUT};
+use crate::{
+    registers::PC,
+    target_info::{StackInfo, TargetInfo},
+    Elf, TIMEOUT,
+};
 
 /// Canary value
 const CANARY_U8: u8 = 0xAA;
@@ -45,6 +49,7 @@ pub struct Canary {
     stack_available: u32,
     data_below_stack: bool,
     measure_stack: bool,
+    size_kb: f64,
 }
 
 impl Canary {
@@ -57,55 +62,27 @@ impl Canary {
         elf: &Elf,
         measure_stack: bool,
     ) -> Result<Option<Self>, anyhow::Error> {
-        let stack_info = match &target_info.stack_info {
-            Some(stack_info) => stack_info,
-            None => {
-                log::debug!("couldn't find valid stack range, not placing stack canary");
-                return Ok(None);
-            }
+        let canary = match Self::prepare(&target_info.stack_info, elf, measure_stack) {
+            Some(canary) => canary,
+            None => return Ok(None),
         };
 
-        if elf.program_uses_heap() {
-            log::debug!("heap in use, not placing stack canary");
-            return Ok(None);
-        }
-
-        let stack_start = *stack_info.range.start();
-        let stack_available = *stack_info.range.end() - stack_start;
-
-        let size = if measure_stack {
-            // When measuring stack consumption, we have to color the whole stack.
-            stack_available as usize
-        } else {
-            round_up(stack_available / 10, 4) as usize
-        };
-
-        log::debug!(
-            "{stack_available} bytes of stack available ({:#010X} ..= {:#010X}), using {size} byte canary",
-            stack_info.range.start(),
-            stack_info.range.end(),
-        );
-
-        let size_kb = size as f64 / 1024.0;
         if measure_stack {
             // Painting 100KB or more takes a few seconds, so provide user feedback.
-            log::info!("painting {size_kb:.2} KiB of RAM for stack usage estimation");
+            log::info!(
+                "painting {:.2} KiB of RAM for stack usage estimation",
+                canary.size_kb
+            );
         }
         let start = Instant::now();
-        paint_subroutine::execute(core, stack_start, size as u32)?;
+        paint_subroutine::execute(core, canary.address, canary.size as u32)?;
         let seconds = start.elapsed().as_secs_f64();
         log::trace!(
             "setting up canary took {seconds:.3}s ({:.2} KiB/s)",
-            size_kb / seconds
+            canary.size_kb / seconds
         );
 
-        Ok(Some(Canary {
-            address: stack_start,
-            size,
-            stack_available,
-            data_below_stack: stack_info.data_below_stack,
-            measure_stack,
-        }))
+        Ok(Some(canary))
     }
 
     /// Detect if the stack canary was touched.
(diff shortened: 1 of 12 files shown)


src/cli.rs
```diff
@@ -123,6 +123,10 @@ pub fn handle_arguments() -> anyhow::Result<i32> {
         }
     });
 
+    if opts.measure_stack {
+        log::warn!("use of deprecated option `--measure-stack`: Has no effect anymore and will vanish on next minor release")
```
jonathanpallant, line 127: I think we have to wait for a major release to take away command line arguments?

src/main.rs
```diff
@@ -71,9 +71,9 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
     let target_info = TargetInfo::new(elf, memory_map, probe_target, stack_start)?;
 
     // install stack canary
-    let canary = Canary::install(core, &target_info, elf, opts.measure_stack)?;
-    if opts.measure_stack && canary.is_none() {
-        bail!("failed to set up stack measurement");
+    let canary = Canary::install(core, &target_info, elf)?;
+    if canary.is_none() {
+        log::info!("stack measurement was not set up");
```
jonathanpallant, line 76: Should this be a warn? Or are there sensible reasons why it might not happen?

jonathanpallant:
Left some notes, but good job! Looks great!

src/canary.rs
```diff
@@ -339,22 +333,19 @@ mod measure_subroutine {
...
+        let word_addr = core.read_core_reg(RegisterId(0))?;
 
         // take a closer look at word, to get address of lowest touched byte
         let offset = core
-            .read_word_32(word_addr as u64)?
+            .read_word_32(word_addr)?
             .to_le_bytes()
             .into_iter()
             .position(|b| b != CANARY_U8)
             .unwrap();
```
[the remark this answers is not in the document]
jonathanpallant: (edit: Github is pointing at a not very helpful line of code - the comment applies to the whole function)

src/cli.rs
Urhengulas: The current version is `v0.3.9`. In my understanding of semver the next breaking release will be `v0.4.0`.

If the version would be e.g. `v1.3.9`, the next breaking release would be `2.0.0`.

Correct me if I am wrong.

src/canary.rs
```diff
@@ -53,142 +63,124 @@ impl Canary {
...
-    /// Detect if the stack canary was touched.
-    pub fn touched(self, core: &mut Core, elf: &Elf) -> anyhow::Result<bool> {
-        let size_kb = self.size as f64 / 1024.0;
-        if self.measure_stack {
-            log::info!("reading {size_kb:.2} KiB of RAM for stack usage estimation");
-        }
+    /// Measure the stack usage.
+    ///
+    /// Returns `true` if a stack overflow is likely.
+    pub fn measure(self, core: &mut Core, elf: &Elf) -> anyhow::Result<bool> {
```
Urhengulas, line 92: I just noticed that we take `self` and therefore consume the object. Is this desired?

At the moment we only ever call this function once, therefore it kinda does not matter.

It does compile with `&self`, but I am wondering if that is a good idea, because it means that this function could be called from multiple threads, without the need to synchronize the accesses. That probably (assumption!) would be bad, because it quite likely will confuse the debugger.

Urhengulas: If it is desired I will add a note stating that.

src/canary.rs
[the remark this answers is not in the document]
Urhengulas: I guess it is possible because we are dealing with hardware here xD

But in that case I'd expect this function to return `0` which then will get reported as `(HOST) INFO  program has used at least 0.0/254.93 KiB (0.0%) of stack space`. Which I think is fair.

src/canary.rs
jonathanpallant: You can't pass the reference between threads unless the type implements the Send trait. If you wish to avoid that, you can add a field that is not Send (like a `PhantomData<*mut ()>` or something)

src/canary.rs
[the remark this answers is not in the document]
jonathanpallant: Returning 0 is reasonable. Unwrapping is probably not reasonable.

src/main.rs
jonathanpallant: No, info is fine.

src/cli.rs
jonathanpallant: That is correct.

jonathanpallant: Which means 0.4 is a "major" release, even though you updated the middle digit.

jonathanpallant: Although semver.org says:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

So technically you can do what you like.

Urhengulas: Interesting. My brain parses it as major.minor.patch, so I though it would be a minor release if we bump the middle number. I will change it to "breaking release".

src/canary.rs
Urhengulas: I will just leave it at `self`. We currently don't want to call it multiple times. If we ever want to, we can make a more informed decision because we know why we would want it.

src/canary.rs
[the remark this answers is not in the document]
Urhengulas: I did change it back to returning an `Option`. And I replaced the `.unwrap()` with an `.expect()`.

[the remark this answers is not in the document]
jonathanpallant: I think I might suggest the other way? Return a u32, but replace the unwrap with an `unwrap_or_else(...)` so you do not panic if the every byte in the stack is equal to the CANARY_U8 but instead cause the function to return "0 bytes of stack used".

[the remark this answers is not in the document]
Urhengulas: The u32 is actually the address and not the amount of memory used. The address is only calculated in `fn measure`. So the function kinda reports if we found a touched address, or if we found none. So the Option kinda makes sense.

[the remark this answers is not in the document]
jonathanpallant: Oh, I see. This particular line of code is looking for the first non-CANARY_U8 byte within a word that has already been checked as being != CANARY_U32. So the unwrap on line 341 is probably appropriate (but could be documented to avoid confusion). Sorry for the brain fog.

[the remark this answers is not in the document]
Urhengulas: No worries, haha. I did confuse it as well.

I tried to make it more clear by adding the error message. Do you think this makes it better?

https://github.com/knurling-rs/probe-run/blob/30f2284d080a4a1d2d57894a3172e6397c6d63d1/src/canary.rs#L380

jonathanpallant (approved):

Issue #412, opened 2023-06-20
println! unexpectedly prints the statement's location before the formatted string


whitequark:
I was looking for a way to get my debug output to align into unbroken tables. Initially I wondered if it's possible to put the log output on a single line, but in https://github.com/knurling-rs/probe-run/issues/407 it is argued that the need to add a second textual log format is not convincingly present, so I guess that's not happening.

Since it is not in `println!`'s contract that it prints a bunch of random text in addition to the formatted string, and the proposal in #407 that would be an imperfect but functional workaround is rejected, I think there needs to be an option to disable log message headers for `println!`.

Issue #407, opened 2023-05-21
Allow single line log output


ctron:
**Describe the bug**

Allow the user to choose a single-line output format.

**To Reproduce**

Use `probe-run` in combination with defmt.

**Expected and observed behavior**

When I run something, I get the following output:

```
0.000274 INFO  Performing reset ...
└─ vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:82
0.701110 INFO  Performing reset ... done!
└─ vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:89
0.701477 INFO  Sending wakup
└─ vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:93
1.702239 INFO  Run
└─ vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:99
1.810455 INFO  Firmware: FirmwareVersion { ic: 50, version: 1, revision: 6, supports_iso18092: true, supports_iso14443_a: true, supports_iso14443_b: true }
└─ vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:107
```

I find this hard to read and would prefer a format with a single line output, like this:

```
0.000274 INFO  Performing reset ...
0.701110 INFO  Performing reset ... done!
0.701477 INFO  Sending wakup
1.702239 INFO  Run
1.810455 INFO  Firmware: FirmwareVersion { ic: 50, version: 1, revision: 6, supports_iso18092: true, supports_iso14443_a: true, supports_iso14443_b: true }
```

Or, some mix:

```
0.000274 [vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:82] INFO  Performing reset ...
0.701110 [vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:89] INFO  Performing reset ... done!
0.701477 [vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:93] INFO  Sending wakup
1.702239 [vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:99] INFO  Run
1.810455 [vat_card_reader::____embassy_main_task::{async_fn#0} @ src/main.rs:107] INFO  Firmware: FirmwareVersion { ic: 50, version: 1, revision: 6, supports_iso18092: true, supports_iso14443_a: true, supports_iso14443_b: true }
```

Urhengulas:
[...]

PR #413, opened 2023-06-20
Hide location printing for defmt frames behind another verbosity level

whitequark:
Depends on https://github.com/knurling-rs/defmt/pull/762.

Addresses #412 and #407.

This PR changes the default (without flags) output to omit location for all defmt frames. Personally I think this is more convenient when you're _running_ your firmware and not _debugging_ it since the log is easier to read and you don't care about locations at this point anyway, but this comes down to personal preference. A less invasive option is to add a `--terse` option and to set the verbosity level to `1 + opts.verbose - opts.terse`; then the default output will remain the same.

A very minimal solution that addresses #412 but not #407 is to skip printing of location only for `println!` frames if `always_include_location` is not set, and leave everything else the same.

I'm happy to implement any of the options above, or really any other option that's suggested.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -107,17 +107,19 @@ pub fn handle_arguments() -> anyhow::Result<i32> {
     let opts = Opts::parse();
     let verbose = opts.verbose;
 
-    defmt_decoder::log::init_logger(verbose >= 1, opts.json, move |metadata| {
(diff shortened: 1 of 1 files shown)


Urhengulas:
Thank you. I will look into the topic

Urhengulas:
Superseded by https://github.com/knurling-rs/probe-run/pull/416

Issue #354, opened 2022-09-25
Support for viewing defmt RTT logs without resetting device


cstrahan:
Hello,

Is there a means (or would be possible to add support) for a way to view `defmt` logs while a micro controller is already running (i.e. without resetting it first)? My current use case looks something like:

* I have a Raspberry Pi Pico
* I can bring the RUN pin low to reset the device
* When I do, sometimes the program resets and starts running from the beginning successfully...
* ... and other times, the device appears to lock up or crash -- the LCD I have attached as part of my project freezes, but I don't know if that's because the LCD controller is stuck in a bad state while everything else on the RP2040 is chugging along, or if something `panic`ed, or...

Logs here would be a big help.

I can do `probe-run target/thumbv6m-none-eabi/debug/pim --no-flash --chip rp2040`, which doesn't flash the device, but it _does_ reset it. It would help in this sort of scenario if I could view the `defmt` logs _without_ resetting the device.

I guess such a feature contradicts the "run" in the name `probe-run`, but unless there's already another executable that provides this ability, it would be really nice to have. Or if this is a fundamental limitation of `defmt`/`probe-run`, it would be nice to have an advisory in the docs (assuming I haven't overlooked such).

Thanks!

cstrahan:
Here's where I've described how to use `defmt` with `cargo-embed`: https://github.com/probe-rs/probe-rs/discussions/1237#discussioncomment-3727155

Still would be cool for `probe-run` to support this, as (at the very least) I really like the unique features `probe-run` has like colorization for the various log components (making it easier to spot log level, file names, etc) and support for full back traces.

whitequark:
https://github.com/knurling-rs/probe-run/pull/414

PR #414, opened 2023-06-21
Implement --no-reset mode for live attach to a running device

whitequark:
Fixes https://github.com/knurling-rs/probe-run/issues/354. Includes some related cleanups.

In my testing this works quite well, including backtraces. The one thing I haven't tested is really flooding defmt with messages, but it probably works fine.


src/cli.rs
```diff
@@ -75,6 +75,10 @@ pub struct Opts {
     #[arg(long, env = "PROBE_RUN_PROBE")]
     pub probe: Option<String>,
 
+    /// Whether to reset the target when attaching and detaching. Implies `--no-flash`.
+    #[arg(long = "no-reset", default_value = "true", action = ArgAction::SetFalse)]
+    pub reset: bool,
```
Urhengulas, line 80: I prefer to keep the field name and the name of the cli option the same. Otherwise it could get confusing.

```suggestion
    #[arg(long)]
    pub no_reset: bool,
```

Please maintain the alphabetical order when you change the name.

src/main.rs
```diff
@@ -54,30 +49,42 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
...
+        // install stack canary
+        canary = Canary::install(core, &target_info, elf, opts.measure_stack)?;
+        if opts.measure_stack && canary.is_none() {
+            bail!("failed to set up stack measurement");
+        }
+    } else {
+        // cannot safely touch the stack of a running application
+        canary = None;
+        log::warn!("You are using `--no-reset` -- stack overflow checks will not be done")
     }
```
Urhengulas, line 84: After #410 the code looks like this:

https://github.com/knurling-rs/probe-run/blob/1b2140b9b87d5cf24bf94c2b201e93ca1d55217c/src/main.rs#L73-L77

Maybe you can change it to something like following
```suggestion
    // install stack canary
    let canary = match opts.no_reset {
        true => None, // cannot safely touch the stack of a running application
        false => Canary::install(core, &target_info, elf, true)?,
    };
    if canary.is_none() {
        log::info!("stack measurement was not set up");
    }
```

src/cli.rs
```diff
@@ -60,7 +60,7 @@ pub struct Opts {
     list_probes: bool,
 
     /// Whether to measure the program's stack consumption.
-    #[arg(long)]
+    #[arg(long, conflicts_with = "no-reset")]
```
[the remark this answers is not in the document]
Urhengulas: Also please rebase.

src/main.rs
```diff
@@ -54,30 +49,42 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
...
-    // reset-halt the core; this is necessary for analyzing the vector table and
-    // painting the stack
-    core.reset_and_halt(TIMEOUT)?;
+    // reset if requested, and then halt the core; this is necessary for analyzing
+    // the vector table and painting the stack
+    if opts.reset {
+        core.reset_and_halt(TIMEOUT)?;
+    } else {
+        core.halt(TIMEOUT)?;
+    }
```
Urhengulas, line 66:
```suggestion
    match opts.no_reset {
        false => core.reset_and_halt(TIMEOUT),
        true => core.halt(TIMEOUT),
    }?;
```

src/main.rs
```diff
@@ -93,8 +100,13 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
         backtrace::Settings::new(canary_touched, current_dir, halted_due_to_signal, opts);
     let outcome = backtrace::print(core, elf, &target_info, &mut backtrace_settings)?;
 
-    // reset the target
-    core.reset_and_halt(TIMEOUT)?;
+    if opts.reset {
```
Urhengulas, line 103:
```suggestion
    if !opts.no_reset {
```

src/main.rs
```diff
@@ -229,16 +231,33 @@ fn start_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
     Ok(())
 }
 
+fn detach_from_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
+    log::debug!("detaching from program");
+
+    if let Some(rtt_buffer_address) = elf.rtt_buffer_address() {
+        set_rtt_blocking_mode(core, rtt_buffer_address, false)?;
```
Urhengulas, line 238: What if rtt was in non-blocking mode before we attached? Then we probably don't want to set it to blocking afterwards., right?

src/main.rs
```diff
@@ -202,24 +214,14 @@ fn flashing_progress() -> flashing::FlashProgress {
...
-    log::debug!("starting device");
+fn attach_to_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
+    log::debug!("attaching to program");
 
     match (core.available_breakpoint_units()?, elf.rtt_buffer_address()) {
         (0, Some(_)) => bail!("RTT not supported on device without HW breakpoints"),
         (0, None) => log::warn!("device doesn't support HW breakpoints; HardFault will NOT make `probe-run` exit with an error code"),
-        (_, Some(rtt_buffer_address)) => set_rtt_to_blocking(core, elf.main_fn_address(), rtt_buffer_address)?,
+        (_, Some(rtt_buffer_address)) =>
+            set_rtt_blocking_mode(core, rtt_buffer_address, true)?,
```
Urhengulas, line 224: Don't we need to halt the program at the beginning of main? Before this PR this was done in `set_rtt_to_blocking`.

src/main.rs
```diff
@@ -229,16 +231,33 @@ fn start_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
...
 }
 
+fn detach_from_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
+    log::debug!("detaching from program");
+
+    if let Some(rtt_buffer_address) = elf.rtt_buffer_address() {
+        set_rtt_blocking_mode(core, rtt_buffer_address, false)?;
+    }
+
+    core.clear_hw_breakpoint(cortexm::clear_thumb_bit(elf.vector_table.hard_fault).into())?;
```
Urhengulas, line 241: Why do we need this line?

src/main.rs
```diff
@@ -202,24 +214,14 @@ fn flashing_progress() -> flashing::FlashProgress {
...
-/// Returns `(stack_start: u32, reset_fn_address: u32)`
-fn analyze_vector_table(core: &mut Core) -> anyhow::Result<(u32, u32)> {
-    let stack_start = core.read_core_reg::<u32>(SP)?;
-    let reset_address = cortexm::set_thumb_bit(core.read_core_reg::<u32>(PC)?);
-    Ok((stack_start, reset_address))
-}
-
-fn start_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
-    log::debug!("starting device");
+fn attach_to_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
```
Urhengulas, line 217:
```suggestion
fn attack_the_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
```
😆

src/main.rs
```diff
@@ -229,16 +231,33 @@ fn start_program(core: &mut Core, elf: &Elf) -> anyhow::Result<()> {
...
-    core.run()?;
-    core.wait_for_core_halted(Duration::from_secs(5))?;
+    log::debug!(
+        "setting RTT mode to {}",
+        if is_blocking {
+            "blocking"
+        } else {
+            "nonblocking"
+        }
+    );
```
Urhengulas, line 260:
```suggestion
    let (mode_name, mode) = match is_blocking {
        true => ("blocking", MODE_BLOCK_IF_FULL),
        false => ("nonblocking", MODE_NON_BLOCKING_TRIM),
    };

    log::debug!("setting RTT mode to {mode_name}");
```

src/main.rs
```diff
@@ -249,14 +268,17 @@ fn set_rtt_to_blocking(
...
     const MODE_MASK: u32 = 0b11;
+    const MODE_NON_BLOCKING_TRIM: u32 = 0b01;
     const MODE_BLOCK_IF_FULL: u32 = 0b10;
-    let modified_channel_flags = (channel_flags[0] & !MODE_MASK) | MODE_BLOCK_IF_FULL;
+    let modified_channel_flags = (channel_flags[0] & !MODE_MASK)
+        | if is_blocking {
+            MODE_BLOCK_IF_FULL
+        } else {
+            MODE_NON_BLOCKING_TRIM
+        };
```
Urhengulas, line 278:
```suggestion
    let modified_channel_flags = (channel_flags[0] & !MODE_MASK) | mode;
```

src/backtrace/mod.rs
```diff
@@ -130,7 +130,7 @@ impl Outcome {
             Outcome::StackOverflow => log::error!("the program has overflowed its stack"),
             Outcome::HardFault => log::error!("the program panicked"),
             Outcome::Ok => log::info!("device halted without error"),
-            Outcome::CtrlC => log::info!("device halted by user"),
+            Outcome::CtrlC => log::info!("interrupted by user"),
```
Urhengulas, line 133:
```suggestion
            Outcome::CtrlC => log::info!("device interrupted by user"),
```

Urhengulas (changes requested):
Looks great. I added a few suggestions and questions.

PS. you are allowed to disagree with the suggestions.

src/cortexm.rs
```diff
@@ -46,8 +46,10 @@ pub fn subroutine_eq(addr1: u32, addr2: u32) -> bool {
 /// The contents of the vector table
 #[derive(Debug)]
 pub struct VectorTable {
-    // entry 0
+    // entry 0: Initial SP
     pub initial_stack_pointer: u32,
+    // entry 1: Reset vector
+    pub reset: u32,
```
Urhengulas, line 52: There was some discussion if we can extract the reset vector from the elf or not and how to make this as stable as possible. Please see https://github.com/knurling-rs/probe-run/pull/383#issuecomment-1438840061. Do you think we can always™️ extract the reset vector the way you implemented it?

whitequark: From the linked discussion:

> It could be that the Flash memory at 0x2000_0000 is aliased at address 0x0000_0000 but I have not checked.

I think that's a typo: the Flash memory is (on STM32s for example; in principe it could be anywhere the instruction fetches are legal) at 0x0800_0000 and the SRAM is (always) at 0x2000_0000.

However, that aside, I looked it up in the ARMv7-M ARM and it seems that I misremembered some details and a much more reliable way would be to read the IVT address from the VTOR register on attach and then use that. I can implement that later.

Screenshot_20230622_171704

src/backtrace/mod.rs
whitequark: I changed it to the current message because it's the *probe* that was interrupted, not the device -- the device keeps running if you do `--no-reset`.

src/main.rs
```diff
@@ -54,30 +49,42 @@ fn run_target_program(elf_path: &Path, chip_name: &str, opts: &cli::Opts) -> any
...
+        }
+    } else {
+        // cannot safely touch the stack of a running application
+        canary = None;
+        log::warn!("You are using `--no-reset` -- stack overflow checks will not be done")
     }
 
     // run program and print logs until there is an exception
-    start_program(core, elf)?;
+    attach_to_program(core, elf)?;
```
whitequark, line 87: Gotta update it here too:
```suggestion
    attack_the_program(core, elf)?;
```

src/main.rs
whitequark: Is there a reason it needs to be? I couldn't find one and that's why I removed the check that was in `set_rtt_to_blocking`.

src/backtrace/mod.rs
Urhengulas: Maybe we can actually have two different messages then?

```suggestion
            Outcome::CtrlC if opts.no_reset == true => log::info!("interrupted by user; program keeps running"),
            Outcome::CtrlC if opts.no_reset == false => log::info!("device halted by by user"),
```

src/main.rs
Urhengulas: Oh no 😆  I am sorry. This was meant to be a joke 🙈  Please change it back to `attach_to_program` 🤣

src/cortexm.rs
Urhengulas: > I can implement that later.

That would be nice. Maybe in a follow up PR.

In any case I we will need to test the current PR with the rp pico, because that one seems to behave a bit differently than many other microcontrollers because of the bootloader.

Urhengulas: I can do the testing. Just need to set it up.

src/main.rs
Urhengulas: I assume (assume!) that we want to wait for potential startup code to finish before we mess around with the micro-controller. Runtime crates like `cortex_m_rt` ([link](https://docs.rs/cortex-m-rt/latest/cortex_m_rt/)) run some code before `main` and also allow users to specify custom code to run before `main`.

src/backtrace/mod.rs
Urhengulas: It's not, but we can just add it as an argument:

```rust
fn log(&self, no_reset: bool)
```

Urhengulas:
@whitequark What is the status? 😄

whitequark:
@Urhengulas I've been taking care of my health. ETA is this weekend.

Urhengulas:
> @Urhengulas I've been taking care of my health

That is very good! 🍀 
Sorry, I did not want to pressure you

Urhengulas:
@whitequark friendly ping 🛎️

whitequark:
I'm doing way better! I think I'll do this Saturday.

Urhengulas:
> I'm doing way better! I think I'll do this Saturday.

That's great to hear!

whitequark:
Sorry, I don't think I'll find time to work on this any time soon as I've not had a chance to use probe-rs in months. Feel free to do with the PR whatever you like.

whitequark (commented):
It looks like some of my responses never went through.

PR #423, opened 2023-09-20
Add better defaults for log format when timestamp is available

### LINKED ISSUE #421: --log-format is inconvenient when only timestamp must be added
**Describe the bug**
Since #416 the default log format is defined by probe-run instead of defmt-decoder, and that default log format does not contain a timestamp. It is not convenient to define a log format from scratch just to add a timestamp.

**To Reproduce**
Define a timestamp with `defmt::timestamp!()`, then use probe-run.

I get a warning that the log format doesn't contain a timestamp:
```
(HOST) WARN  `defmt::timestamp!` implementation was found, but timestamp is not part of the log format; consider adding the timestamp `{t}` argument to the log format
```

I would expect a timestamp to be automatically used as it was the case before. This dynamic check should be easy based on the presence of `defmt::timestamp!()` which probe-run is able to do because probe-run unilaterally forces a default for log-format (which means defmt decoder taking an optional log-format is probably dead code).

An alternative solution would be to provide a convenience flag (eg. `--log-timestamp`) to use the timestamp version of the log-format (i.e. the default log format prefixed by `{t} `).

andresovela:
Fixes #421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+- [#423] Add better defaults for log format when timestamp is available
+
+[#423]: https://github.com/knurling-rs/probe-run/pull/423
+
 ## [v0.3.10] - 2023-08-01
 
 - [#417] Release `probe-run v0.3.10`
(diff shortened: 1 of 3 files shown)


Urhengulas (approved):
Nice! Tested it locally and it works as expected

Thank you!!

PR #425, opened 2023-10-03
Remove default format logic

andresovela:
This PR depends on knurling-rs/defmt#783.


andresovela:
Since `probe-run` is being replaced by `probe-rs`, I'm closing this PR and make the changes in `probe-rs` instead.

PR #431, opened 2024-01-29
Prepare `v0.3.11` (the final release)

### LINKED ISSUE #430: release final version with deprecation notice
while the deprecation notice is clearly visible in the README on _github_ it is not visible on crates.io and will thus probably be missed by many (by all appearances on crates.io it's still the correct tool to use).
it'd be great if you could release a final v0.30.1 to crates.io just to get the README published there.

optionally you could maybe even print a deprecation notice to STDOUT (or STDERR?) on each application invocation so that those which just update it via `cargo install` (or more likely maybe [`cargo install-update`](https://crates.io/crates/cargo-update)) and don't check crates.io either also see it.

with the release (if you publish it also on GH, as you've done so far) people will also see it in their GH feed (if they look at it), spreading the information even further.

i just stumbled across the announcement by chance and would otherwise have missed it completely.

Urhengulas:
Using the `deprecated` attribute is a bit of a hack, but it was the easiest way to issue a warning on build time I could find.

Fixes #430 

---

TODO
- [x] publish release (`cargo publish`)
- [x] add `v0.3.11` git tag
- [ ] archive repository

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+## [v0.3.11] - 2024-01-29
+
 - [#423] Add better defaults for log format when timestamp is available
 
 [#423]: https://github.com/knurling-rs/probe-run/pull/423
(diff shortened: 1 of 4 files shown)


jonathanpallant:
Makes sense to me.

Can you post a sample of running the binary with this new deprecated function in it?

Urhengulas:
@jonathanpallant said:
> Can you post a sample of building / running the binary with this new deprecated function in it?

```console
$ cargo install --path . --debug -q
warning: use of deprecated function `deprecated`: ⚠️  As of 11.10.2023 `probe-run` is in maintainance mode. We recommend everyone to switch to `probe-rs`. Read following article on the why and on how to migrate: https://ferrous-systems.com/blog/probe-run-deprecation/
  --> src/main.rs:53:5
   |
53 |     deprecated();
   |     ^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

$ probe-run
⚠️  As of 11.10.2023 `probe-run` is in maintainance mode. We recommend everyone to switch to `probe-rs`. Read following article on the why and on how to migrate: https://ferrous-systems.com/blog/probe-run-deprecation/

error: the following required arguments were not provided:
  --chip <CHIP>
  <ELF>

Usage: probe-run --chip <CHIP> <ELF>

For more information, try '--help'.
```

jonathanpallant:
I was going to suggest adding a flag to turn the warning off, but actually I'd suggest they just fork the codebase and take the warning out if they care that much about some probe-run feature that probe-rs doesn't have.

https://github.com/ferrous-systems/rust-exercises/pull/66 is in, so this all seems fine to me.

jonathanpallant (approved):

